Issue #16, opened 2017-04-12
comparison to `auditwheel`


johnyf:
The `delocate` package appears to be the OS X counterpart of [`auditwheel`](https://github.com/pypa/auditwheel), which is for Linux (and seems to have been a more recent development than `delocate`). It could be helpful to point to `auditwheel` from the README, in case one is interested in amending Linux wheels instead, or Linux wheels too. It may also be useful to briefly compare to `auditwheel` in terms of functionality.

matthew-brett:
Right - actually auditwheel was half-fork of delocate and includes some delocate code.

Any interest in doing a quick PR to add the link in the README?

johnyf:
Thank you for the explanation, I had not noticed https://github.com/pypa/auditwheel/commit/4562a2928cc6f20d162e102b53b950f8768e91ca. I opened a PR that adds the link to `auditwheel`.

PR #17, opened 2017-04-15
DOC: link to `auditwheel` from README

johnyf:
Addresses #16.

The only way I know of formatting text within a link as code is

```rst
`:code:`auditwheel` <https://github.com/pypa/auditwheel>`_
```

This works locally (with `rst-preview-pandoc@0.1.12` in `atom`), but doesn't render properly on GitHub, so I let the package's name be plain text.


matthew-brett:
Great - thanks.

PR #18, opened 2017-06-02
Allow for xcode command line tools names in output

sgillies:
Trailing periods also may or may not exist in xcode messages.

Testing for strings within xcode messages will be more reliable and future proof than testing that xcode messages have specific beginnings or endings.

I needed this patch to delocate macosx wheels using XCode 8.2.1.


matthew-brett:
Thanks - looks reasonable.  I'll let the tests run and then merge.

PR #19, opened 2017-06-04
Replace invalidated signatures in delocated libraries.

HexDecimal:
If a binary file is signed, tools like install_name_tool will invalidate
the signature by changing the file contents.  Later versions of MacOS
refuse to load binary files with invalid signatures.

This adds a new function which overwrites bad signatures with an ad-hoc
signature.  This is as close as you can get to removing the signature
using the available tools in MacOS.

### DIFF
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -437,3 +437,44 @@ def lipo_fuse(in_fname1, in_fname2, out_fname):
     return back_tick(['lipo', '-create',
                       in_fname1, in_fname2,
                       '-output', out_fname])
+
+
+@ensure_writable
+def replace_signature(filename, identity):
+    """ Replace the signature of a binary file using `identity`
+
+    See the codesign documentation for more info
+
+    Parameters
+    ----------
+    filename : str
+        Filepath to a binary file.
+    identity : str
+        The signing identity to use.
+    """
+    back_tick(['codesign', '--force', '--sign', identity, filename],
+              raise_err=True)
+
+
+def validate_signature(filename):
+    """ Remove invalid signatures from a binary file
+
+    If the file signature is missing or valid then it will be ignored
+
+    Invalid signatures are replaced with an ad-hoc signature.  This is the
+    closest you can get to removing a signature on MacOS
+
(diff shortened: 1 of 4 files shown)


HexDecimal:
Added a wheel test to this PR.

I probably over-engineered this.  I can remove the `validate_signature` function if you don't care about leaving signature-less files untouched.

delocate/tools.py
```diff
@@ -437,3 +437,43 @@ def lipo_fuse(in_fname1, in_fname2, out_fname):
...
+    Parameters
+    ----------
+    filename : str
+        Filepath to a binary file
+    """
+    out, err = back_tick(['codesign', '--verify', filename],
+                         ret_err=True, as_str=True, raise_err=False)
+    if 'code object is not signed at all' in err:
+        return # File has no signature, and adding a new one isn't necessary
+    if not err:
```
matthew-brett: Maybe reverse these clauses for clarity?

delocate/tools.py
```diff
@@ -437,3 +437,43 @@ def lipo_fuse(in_fname1, in_fname2, out_fname):
...
+        return # The existing signature is valid
+    # This files signature is invalid and needs to be replaced
```
matthew-brett: File's

HexDecimal:
I put `if not err:` first and fixed the spelling.

> what is the advantage of ad-hoc signing compared to no signature, or an invalid one?

No signature means there's nothing to break when delocate modifies the file. `validate_signature` was designed to not touch these, since it's hard to remove signatures once they're added.

There isn't an official tool to remove bad signatures so I replace them with an ad-hoc signature.  The only documentation I could find on ad-hoc was vague.  My guess is that it's just a checksum.

Here's an issue caused by an invalid signature: https://github.com/HexDecimal/python-tdl/issues/26
It raises an OSError, specifically rejecting the dependencies because of signature issues.  I couldn't repeat this on an older MacOS system which loaded the 'invalid' files just fine.

matthew-brett:
OK - thanks - let's see what the public thinks.

PR #20, opened 2017-06-11
Delocate `@rpath` dependencies.

### LINKED ISSUE #11: Dependencies beginning with @rpath
The documentation has the following warning:

> If you see any dependencies starting with @ in a wheel that has not yet been processed with delocate-wheel, the delocation process below will likely not work, and leave you with - er - a broken wheel.

I've just built a cython extension that links to a dynamic library, `libgeos_c.dylib`. This seems to result in the .so being linked using `@rpath`:

```
(python34)snorfmac:Shapely snorf$ otool -L build/lib.macosx-10.5-x86_64-3.4/shapely/speedups/_speedups.so 
build/lib.macosx-10.5-x86_64-3.4/shapely/speedups/_speedups.so:
    @rpath/./libgeos_c.1.dylib (compatibility version 10.0.0, current version 10.2.0)
    /usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
```

What do I need to do in this case? Why has the linker chosen to do this? How do I tell it not to so I can delocate the wheel? As far as I'm aware there is nothing unusual in `setup.py`.

I'm running OS X 10.11.3, using Anaconda Python (3.4.4).

There are several instances of the library on the system - I wonder if this is an issue?

```
(python34)snorfmac:Shapely snorf$ mdfind -name libgeos_c.1.dylib
/Users/snorf/Desktop/tidy/Shapely/cxfreeze_example/build/exe.macosx-10.10-x86_64-3.4/libgeos_c.1.dylib
/usr/local/Cellar/geos/3.5.0/lib/libgeos_c.1.dylib
/Users/snorf/miniconda3/pkgs/geos-3.4.2-0/lib/libgeos_c.1.dylib
/Users/snorf/miniconda3/envs/python34/lib/libgeos_c.1.dylib
/usr/local/Cellar/geos/3.4.2/lib/libgeos_c.1.dylib
/usr/local/lib/python3.4/site-packages/shapely/libgeos_c.1.dylib
```

HexDecimal:
Added `delocating.resolve_rpath` function.  This function needs
`copy_flit_func` which is now passed in `delocating.delocate_path`.

`tools.get_rpaths` changed to handle non-library inputs.

delocate_wheel no longer changes directory.  This caused issues
when library dependencies had relative paths without `@`.

Resolves #11

I don't have an easy way to make a test wheel for this.  I only have a minimal test for the `resolve_rpath` function.


delocate/libsana.py
```diff
@@ -51,16 +58,71 @@ def tree_libs(start_path, filt_func = None):
...
                     lib_dict[lib_path] = {depending_libpath: install_name}
     return lib_dict
 
 
+def resolve_rpath(lib_path, rpaths, copy_filt_func):
+    """ Return `lib_path` with its `@rpath` resolved
+
+    If the `lib_path` does't have `@rpath` then it's returned as is.
+
+    If `lib_path` has `@rpath` then returns the first `rpaths`/`lib_path`
```
matthew-brett: I'm not sure this logic is right.  Imagine a library that depends on `@rpath/foo.dylib`.  The rpath contains `/path1, /path2` both of which contain `foo.dylib`. `copy_filt_func` skips `/path1/foo.dylib` for some reason.  Now the returned lib path is `/path2/foo.dylib` even though this is not in fact the `foo.dylib` that the library will link to at run time.

I think `copy_filt_func` is saying, "don't copy stuff from here", but not "don't link to stuff from here".  If you see what I mean.

HexDecimal:
Fixed the typo and removed the use of `copy_filt_func`.

I'll try to set up a wheel, but I don't have a local MacOS environment to build it on.

y0ast:
I have access to a local MacOs, but not 10.6.

I can probably still help test it @HexDecimal.

HexDecimal:
I've added a wheel with a signed `@rpath` dependency.  It's only built by Travis right now so it'll be missing from the `test/data` directory unless delocate is deployed from Travis or somebody runs the `make_wheels.sh` script and uploads the new files.

delocate/tools.py
```diff
@@ -262,7 +264,12 @@ def get_rpaths(filename):
     rpath : tuple
         rpath paths in `filename`
     """
-    out = back_tick(['otool', '-l', filename])
+    try:
+        out = back_tick(['otool', '-l', filename])
+    except RuntimeError:
```
matthew-brett: Can you explain this clause?  Shouldn't it raise an error, as before?

HexDecimal: When I added `get_rpaths` before the `get_install_names` function in `tree_libs` I needed it to ignore non-binary files like `get_install_names` does.  This `try..except` clause can be moved to `tree_libs` if you like.

matthew-brett: No, it's fine as is thanks.

y0ast:
Unfortunately, I have a very hard time replicating the exact build environment for the wheels as used by @matthew-brett.

I could have another stab, but if @matthew-brett has access to the old environment used for creating the other wheels that would be great :).

HexDecimal:
> How easy would it be to make a test for the behavior of skipping missing libraries on the rpath, and choosing the first existing library on the rpath?

I've now added tests for "not found on the first path" and "not found on any path" to `test_resolve_rpath`.

matthew-brett:
OK - thanks - let's give it a try ...

y0ast:
@matthew-brett would you mind making another release that contains this feature?

matthew-brett:
Done just now - 0.7.0.

PR #26, opened 2017-11-13
Expand the rpath search paths to include DYLD_LIBRARY_PATH and DYLD_FALLBACK_LIBRARY_PATH as well

### LINKED ISSUE #24: Allow rpath fallback to LD_LIBRARY_PATH
As I understand it, if the linker can't find a particular library relative to an entry in `@rpath`, then it will fall back to searching other locations, starting with `LD_LIBRARY_PATH`.  `auditwheel` supports a similar approach when trying to find dependencies, and it makes it a whole lot easier to find and include third-party libraries. Is there any reason this sort of thing couldn't be added to `delocate`? Happy to send a pull request your way.

fbrennen:
Fixes #24


fbrennen:
Seems like travis had an unexpected failure on one of those jobs -- any chance you could restart it?

HexDecimal:
Looks like LD_LIBRARY_PATH was `''` which is causing the build failure because `if ld_path is not None:` is True in that case.

Check for empty strings after splitting.  This code will filter them all out:

    for path in os.environ.get('LD_LIBRARY_PATH', '').split(':'):
        if path:
            paths.append(path)

fbrennen:
Yeah, that was definitely the case in the first set of builds, but not the second as far as I can tell -- it appears to be failing in `get_macpython_environment()` because pip wasn't found.
```
++sudo pip uninstall -y virtualenv
sudo: pip: command not found
```
Could be I'm just bad at reading the logs, but all the other builds are passing!

HexDecimal:
You're right, I was looking at the wrong job.  Travis does occasionally hiccup, so the build likely just needs resetting.  You can add your own fork to TravisCI if you want to be certain it passes.

fbrennen:
@matthew-brett it seems like the travis image for xcode8.3 might be a bit borked -- I've replicated the above `pip: command not found` issue when running it on my own. Not sure what the best way to address that is -- does the multibuild script need to be updated, or has something odd happened with the test environment?

fbrennen:
Hah, nevermind, should have poked around a bit more -- I see you have it under control.

fbrennen:
Hi @matthew-brett -- tests are all good now. =)

matthew-brett:
@xoviat - sorry to ask you - but do you have any feelings on this one (it will affect multibuild)?

xoviat:
Seems okay. We really should use nightly delocate for multibuild testing though. It's difficult to anticipate all the things that can go wrong.

fbrennen:
Hope you had a nice holiday -- this seem good to go now?

matthew-brett:
I'm really sorry to be slow.  Yes, I'll merge it soon.

What do you think about DYLD_LIBRARY_PATH, DYLD_FALLBACK_LIBRARY_PATH

Can you think of a good test?  I guess something like this should work for this PR, not otherwise:

```
cd delocate/tests
mkdir foo
# Copy the test application, that depends on liba.dylib
cp data/libb.dylib foo
# It is now broken, because it can't find libc.dylib
# Fix with LD_LIBRARY_PATH
export LD_LIBRARY_PATH=$PWD/data
# Prepend rpath to lib pointer, to pick up LD_LIBRARY_PATH
install_name_tool -change liba.dylib @rpath/liba.dylib foo/libb.dylib
# This should now work.
delocate-path foo
```

fbrennen:
No worries about being slow -- lots of things to do! =)

The test sounds good -- could perhaps get it to work without needing `install_name_tool` as well. I'll have a look.

There is some [developer documentation](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryUsageGuidelines.html#//apple_ref/doc/uid/TP40001928-SW12) that states the paths are searched  LD -> DYLIB -> DYLIB_FALLBACK, so it could make sense to add them all to the search path in that order. I'll get that in too.

matthew-brett:
Hmm - I found in my own tests that I needed to add `@rpath/` to the path in order for LD_LIBRARY_PATH to be effective.  This wasn't true for the DYLD_ variables.  I haven't seen that documented, but I didn't search very hard.

Have a look at `test_delocating.py` for some tests doing similar kinds of things in Python (rather than shell, as in my example).

fbrennen:
Cool, I'll have a look.

fbrennen:
Alright, really sorry about the delay -- how does this look?

matthew-brett:
Did you find, as I did, that you do not need the `rpath` prefix when loading libraries from `DYLD_LIBRARY_PATH, DYLD_FALLBACK_LIBRARY_PATH`.  Does delocation work for that case (no rpath prefix, library in `DYLD_LIBRARY_PATH, DYLD_FALLBACK_LIBRARY_PATH` ?  Should it, do you think?

fbrennen:
I did find that, yes -- it looks like `DYLD` and `DYLD_FALLBACK` also get used to resolve `loader_path` as well. Delocation does not appear to work when `DYLD` paths are set and there is not an `rpath` prefix to a library, which makes sense as all this PR tries to do is enhance `rpath` searching -- nothing else in `delocate` does general `PATH` searching as far as I can tell. 

It would certainly be possible to get that to work, where every relative-path library was searched for on `DYLD_LIBRARY_PATH` and `DYLD_FALLBACK_LIBRARY_PATH`. I can see the logic of that solution, though I don't quite understand why you'd want to link anything using that approach (a relative path without using any relocatable  `@` notation, and libraries hidden inside `DYLD` paths).

matthew-brett:
The justification for using DYLD etc path would be to replicate the library discovery active at the time that delocate runs.  For example, if you set these, then installed the undelocated wheel, it would run, so you could argue that delocate should replicate this.

At the moment it's a bit odd, in that we are replicating this behavior if there's an `rpath` prefix, but not otherwise, even though the `rpath` prefix does not usually influence the use of these variables.  

Does that make sense?

fbrennen:
Sure, totally. I don't mind adding that in.

fbrennen:
Awright, I've been reading through documentation, and what Apple provides doesn't quite match reality. Specifically, here:

https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryUsageGuidelines.html#//apple_ref/doc/uid/TP40001928-SW10

The documentation states that when a library name is a filename then it is first searched for on `LD_LIBRARY_PATH`. I cannot get this to work using the library-name-only test files in `delocate` (e.g. by just moving one into a sub-folder) -- they only get successfully resolved using `DYLD_` paths (which I believe is what you were seeing as well). So either I'm missing something or Apple's approach here has changed.

Thinking about how to address this, it would be simple to just ignore that first case, and always search for non-`@` libraries in the second order specified in the above document: first in `DYLD_LIBRARY_PATH`, then in `realpath`, then in `DYLD_FALLBACK_LIBRARY_PATH`. This could also be applied to libraries using `@loader_path` (as we know what `@loader_path` resolves to -- it's the location of the base library we're resolving the references of). What do you think? 

I'm uncertain about what to do with `@executable_path` as it will be hard to figure out how it could be defined when just loading a library (at least it seems that way to me, unless there's a major overhaul of how `delocate` resolves references). We should still be able to search for those libraries in `DYLD` and `DYLD_FALLBACK`, and just ignore the mystery case where we actually try and resolve `@executable_path`.

fbrennen:
Hokay, here's part of the solution -- what do you think? I had a go at resolving `@loader_path` but ran into an issue: when you first run `delocate-path` all the libraries that are successfully delocated into a subfolder are referred to using `@loader_path`. When you then run `delocate-path` a second time and expand `@loader_path` then `delocate` will no longer skip trying to resolve those library references (because they don't begin with `@` anymore), and that then produces conflicts because the destinations for the delocated libraries (in the subfolder) already exist. I can think of a few ways of resolving this but wanted to see if you thought it was necessary first -- I could see just completely ignoring resolutions for `@loader_path` instead.

delocate/libsana.py
```diff
@@ -104,6 +107,55 @@ def resolve_rpath(lib_path, rpaths):
...
+    """ Search common environment variables for `lib_path`
+
+    Apple's documentation about how to do this does not quite match what we see
+    in the wild, so we'll use a single approach here:
+
+        1. Search for the basename of the library on DYLD_LIBRARY_PATH
+        2. Search for ``realpath(lib_path)``
+        3. Search for the basename of the library on DYLD_FALLBACK_LIBRARY_PATH
+
+    Testing has suggested that LD_LIBRARY_PATH is not actually used in this
```
matthew-brett: Is that also true for binaries compiled with gcc?  I guess so.

delocate/libsana.py
```diff
@@ -104,6 +107,55 @@ def resolve_rpath(lib_path, rpaths):
...
+    -------
+    lib_path : str
+        Full path of ``basename(lib_path)``'s location, if it can be found, or
+        ``realpath(lib_path)`` if it cannot.
+    """
+    lib_basename = basename(lib_path)
+    potential_library_locations = []
+
+    # 1. Search on DYLD_LIBRARY_PATH
+    DYLD_LIBRARY_PATH = os.environ.get('DYLD_LIBRARY_PATH')
```
matthew-brett: Simplify with something like:

```
def _paths_from_var(varname, lib_basename):
    var = os.environ.get(varname)
    if var is None:
        return []
    return [pjoin(path, lib_basename) for path in var.split(':')]
```

?

delocate/libsana.py
```diff
@@ -55,9 +55,12 @@ def tree_libs(start_path, filt_func=None):
                 continue
             rpaths = get_rpaths(depending_libpath)
             for install_name in get_install_names(depending_libpath):
-                lib_path = (install_name if install_name.startswith('@')
-                            else realpath(install_name))
-                lib_path = resolve_rpath(lib_path, rpaths)
+                # If the library starts with '@rpath' we'll try and resolve an rpath
```
matthew-brett: Install name can also start with `@loader_path`

https://matthew-brett.github.io/docosx/mac_runtime_link.html

Presumably we don't want to run this code in that case?  `if install_name.startswith('@rpath`)` ?

Sorry if I forgot - but did you conclude that the system searches for libs via the environment variables, when the install name has @rpath prefix?

delocate/tools.py
```diff
@@ -320,6 +322,15 @@ def get_rpaths(filename):
         assert cmdsize.startswith('cmdsize ')
         paths.append(RPATH_RE.match(path).groups()[0])
         line_no += 2
+    # We'll search the extra library paths in a specific order:
```
matthew-brett: I'd prefer to have the `get_rpaths` function return the rpaths from the lib, and then, in `tree_libs`, concatenate the results from here, with the results from the environment variables.  OK with you?

delocate/libsana.py
fbrennen: `resolve_rpath()` does its own check to see if `install_name` starts with `'@rpath'` -- it would absolutely work to move that up here if you like. I had been trying to keep the code the way you had it before (where you only checked to see if there was an `@` -- see line 58).

It's been a little while since I investigated it, but I did find (https://github.com/matthew-brett/delocate/pull/26#issuecomment-369306639) that the system will use `DYLD` and `DYLD_FALLBACK` to resolve **@rpath** when searching for libraries.

delocate/libsana.py
fbrennen: Hm, I'm not sure -- I'll try and check. I can't remember off the top of my head if I was using clang or gcc.

delocate/tools.py
fbrennen: Yeah, no prob.

delocate/libsana.py
fbrennen: Sounds good!

delocate/libsana.py
fbrennen: Hokay, make it a bit more explicit about what's happening. I've also double-checked and confirmed that the system uses the three environment variables as search options for `@rpath`.

delocate/libsana.py
matthew-brett: Did you get round to checking this?   The ideal would be some test snippets to put in repository, but outside the main code, to confirm the behavior.

delocate/libsana.py
```diff
@@ -104,6 +113,56 @@ def resolve_rpath(lib_path, rpaths):
...
+        var = os.environ.get(varname)
+        if var is None:
+            return []
+        return [pjoin(path, lib_basename) for path in var.split(':')]
+
+    # 1. Search on DYLD_LIBRARY_PATH
+    potential_library_locations += _paths_from_var('DYLD_LIBRARY_PATH',
+                                                   lib_basename)
+
+    # 2. Search for realpath(lib_path)
```
matthew-brett: Add comment here to say why `realpath` is necessary?

delocate/tests/test_delocating.py
```diff
@@ -507,3 +507,150 @@ def test_bads_report():
                   set([(LIB64A, LIBBOTH, ARCH_32),
                        (LIB64, LIB64, ARCH_32, ARCH_64),
                        (LIB32, ARCH_32)]))
+
+
+def test_ld_library_path_lookups():
```
matthew-brett: I'm puzzled - aren't you ignoring LD_LIBRARY_PATH?

delocate/tests/test_delocating.py
```diff
@@ -507,3 +507,150 @@ def test_bads_report():
...
+    # Test that LD_LIBRARY_PATH can be used to find libs during
+    # delocation
+    old_LD_LIBRARY_PATH = os.environ.get('LD_LIBRARY_PATH', None)
+    try:
+        # Clear out LD_LIBRARY_PATH so we can run in isolation
+        if old_LD_LIBRARY_PATH is not None:
+            del os.environ['LD_LIBRARY_PATH']
+        with InTemporaryDirectory() as tmpdir:
```
matthew-brett: Any way of sharing code between the test functions?  There seems to be a lot of repeated code here.

delocate/tests/test_delocating.py
```diff
@@ -507,3 +507,150 @@ def test_bads_report():
...
+                             '@rpath/{}'.format(os.path.basename(LIBB)))
+            # confirm delocate_path doesn't work
+            delocate_path('subtree', 'deplibs')
+            assert_raises(RuntimeError, back_tick, [test_lib])
+            # Update LD_LIBRARY_PATH and confirm that we can now
+            # successfully delocate test_lib
+            os.environ['LD_LIBRARY_PATH'] = hidden_dir
+            delocate_path('subtree', 'deplibs')
+            back_tick(test_lib)
+    finally:
```
matthew-brett: How about a little context manager like:

```
from contextlib import contextmanager

@contextmanager
def in_env(*env_vars):
    old = {v: os.environ.get(v) for v in env_vars}
    with InTemporaryDirectory() as tmpdir:
        yield
    for v in env_vars:
        val = old[v]
        if val is None:
            del os.environ[v]
        else:
            os.environ[v]  = val

```

fbrennen: That's a much nicer way of doing it!

delocate/tests/test_delocating.py
fbrennen: Yeah, I didn't follow up with a comment in the main thread bringing the PR to your attention yet because I'm going through and re-verifying everything, as it's definitely a bit stale for me (sorry!). I'll put in a script you can run yourself to verify the way things work.

delocate/libsana.py
fbrennen: Agreed -- I'll put a script in `tests/data` that walks thorugh this, like you have with `test-libs`.

With reference to your original comment, I'm not sure I completely understand it -- why would you expect the compiler to change the way that `@rpath` entries are resolved? Those entries should be resolved at run-time, not compile-time.

fbrennen: Ok, so I've done some testing and put a bit of a demonstration in the `testing_osx_rpath_env_variables.sh` script above. I didn't quite have this right: if the library that's being searched for has a directory in its path (which seems to me like it's usually going to be the case, unless it's being referenced by something that's in the same folder as you're delocating things in), then the search order is like I laid out above: check DYLD, then try the full path to the lib, and then try DYLD_FALLBACK. If it *doesn't* have a directory in its path, then you check `LD_LIBRARY_PATH` first, then DYLD, then you search for the lib in the directory you're currently in, and then you search DYLD_FALLBACK. I could add that here as an option if you like.

delocate/libsana.py
fbrennen: See my comment above and the `testing_osx_rpath_env_variables.sh` script.

delocate/tests/test_delocating.py
fbrennen: Ok, so this ended up turning into two different features: 

Originally this PR was about just searching for additional libraries using environment variables, and using those to resolve `@rpath` entries. The choice of environment variables was somewhat arbitrary at the time, and I suggested using all three of the choices. That's where this test came from.

After that, you suggested, quite reasonably, that we should extend library resolution more generally to using environment variables, and I tried to do something a bit more formal like Apple has defined in their developer documentation. This has turned out to be a bit complicated as you'll see from looking at that script above, and it turns out there aren't very many cases where `LD_LIBRARY_PATH` ends up getting used (which is why I suggested ignoring it in `search_environment_for_lib()` above).

It seems to me that it would be less likely that you'd run `delocate_xxx()` from within the folder you want to fix libraries in (i.e. you would normally run `delocate_wheel.py <path>/<to>/<wheel>` and not `delocate_wheel.py ./<wheel>`). That means you'd basically never use `LD_LIBRARY_PATH`, as the path(s) to the libraries being delocated will contain a folder in it / them. Why don't I just get rid of that as an option and only support `DYLD_` paths? It will keep everything simple, and searching for libraries will be the same everywhere then (unlike now, where `LD_LIBRARY_PATH` is used when resolving `@rpath` entries and nowhere else).

matthew-brett:
Yes, that sounds reasonable to me ...

delocate/tests/test_delocating.py
fbrennen: Fixed with your nice context manager suggestion. =)

delocate/tests/test_delocating.py
fbrennen: Right, I've stripped out all the LD_LIBRARY_PATH use -- things should be a lot more consistent now.

delocate/libsana.py
fbrennen: Went for just stripping out `LD_LIBRARY_PATH` to simplify things.

fbrennen:
Right, I think I've addressed the issues you raised -- what do you reckon?

delocate/tests/test_delocating.py
```diff
@@ -2,14 +2,15 @@
 
 from __future__ import division, print_function
 
+from contextlib import contextmanager
```
matthew-brett: Are you using contextmanager in this function?

delocate/tools.py
```diff
@@ -290,7 +290,9 @@ def set_install_id(filename, install_id):
 RPATH_RE = re.compile(r"path (.*) \(offset \d+\)")
 
 def get_rpaths(filename):
-    """ Return a tuple of rpaths from the library `filename`
+    """ Return a tuple of rpaths from the library `filename`, with an
```
matthew-brett: This extra part of the comment doesn't apply to this function, does it?

delocate/libsana.py
```diff
@@ -48,16 +49,24 @@ def tree_libs(start_path, filt_func=None):
...
+            search_paths = rpaths + env_var_paths
             for install_name in get_install_names(depending_libpath):
-                lib_path = (install_name if install_name.startswith('@')
-                            else realpath(install_name))
-                lib_path = resolve_rpath(lib_path, rpaths)
+                # If the library starts with '@rpath' we'll try and resolve it
+                # We'll do nothing to other '@'-paths
+                # Otherwise we'll search for the library using env variables
+                if install_name.startswith('@rpath'):
+                    lib_path = resolve_rpath(install_name, search_paths)
```
matthew-brett: Sorry - just a final check - it's true, then that if the name is `@rpath/foo.dylib`, macOS will search not only the rpaths specified in the list of rpaths from the binary, but also paths on `DYLD_LIBRARY_PATH` etc?

delocate/libsana.py
```diff
@@ -104,6 +113,63 @@ def resolve_rpath(lib_path, rpaths):
     return lib_path
 
 
+def search_environment_for_lib(lib_path):
```
matthew-brett: And this differs from `resolve_rpath` because we can use the basename to search for the library in the specified paths?  So we can find a library specified as`/foo/bar/baz.dylib` at `/path/baz.dylib` where `/path` is in, say, `DYLIB_LIBRARY_PATH`?  Please forgive my slowness.

fbrennen: Just so -- in your example, if `/foo/bar/baz.dylib` exists, and there is also a second `baz.dylib` at `/path/baz.dylib`, then by specifying `DYLD_LIBRARY_PATH=/path` you'll find the `/path` library before the `/foo/bar` one. Check out `test_dyld_library_path_beats_basename()` below.

delocate/tests/test_delocating.py
fbrennen: Er, oops! One moment. =)

delocate/tools.py
fbrennen: Quite right.

delocate/libsana.py
matthew-brett: Sorry - just for my reassurance - if required install name is a full absolute path, such as `/foo/bar/baz.dylib`, rather than a relative path, `baz.dylib`, macOS will nevertheless find `/path/baz.dylib` (where `/path` is in `DYLD_LIBRARY_PATH`) before `/foo/bar/baz.dylib`?

fbrennen: Let me go check that specifically -- I can add it to the test script afterwards (assuming it does in fact work as you describe, which it should!).

fbrennen: Right, I've updated `testing_osx_rpath_env_variables.sh` with a demonstration of this working just like you describe. Give it a go. =)

delocate/libsana.py
```diff
@@ -104,6 +113,63 @@ def resolve_rpath(lib_path, rpaths):
...
+    Returns
+    -------
+    lib_path : str
+        Full path of ``basename(lib_path)``'s location, if it can be found, or
+        ``realpath(lib_path)`` if it cannot.
+    """
+    lib_basename = basename(lib_path)
+    potential_library_locations = []
+
+    def _paths_from_var(varname, lib_basename):
```
matthew-brett: Can't this be at the top level?  It's not using any variables from the enclosing function,  is it?

fbrennen: Sure, it absolutely could be -- I put it in here as it was only really used inside this function, but not a problem to move it out.

matthew-brett:
OK - thank you for the patience - in it goes.

fbrennen:
Hah, thank you for putting up with me for the nearly two years this has taken. =)

PR #27, opened 2017-12-01
BF: add fixes for latest otool

matthew-brett:
Latest otool (900) has different not-object error messages.

Refactor routines reading otool output to deal with these error
messages.  Refactor get_rpaths to use the more general machinery.

### DIFF
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -130,22 +130,36 @@ def parse_install_name(line):
     return IN_RE.match(line).groups()
 
 
+# otool -L strings indicating this is not an object file. The string changes
+# with different otool versions.
+BAD_OBJECT_STRINGS = [
+    'is not an object file',  # otool version cctools-862
+    'The end of the file was unexpectedly encountered',  # cctools-862 (.ico)
+    'The file was not recognized as a valid object file.',  # cctools-895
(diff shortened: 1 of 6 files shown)


matthew-brett:
@HexDecimal - what do you think of the rpath changes?

delocate/tools.py
```diff
@@ -130,22 +130,36 @@ def parse_install_name(line):
...
+def _cmd_out_err(cmd):
+    # Run command, return stdout or stderr if stdout is empty
+    out, err = back_tick(cmd, ret_err=True)
+    out = err if not len(out) else out
+    return out.split('\n')
+
+
 def _line0_says_object(line0, filename):
     line0 = line0.strip()
+    for candidate in BAD_OBJECT_STRINGS:
```
thewtex: @matthew-brett Thanks, this should be more robust :+1:

HexDecimal:
@matthew-brett - I think the updated code looks pretty good.

Is the latest otool program actually being tested?  TravisCI has a new xcode image so it might be a good idea to add `osx_image: xcode9.1` into `.travis.yml`

matthew-brett:
@HexDecimal - thanks - good suggestion - added.

PR #28, opened 2017-12-03
BF: allow for Xcode 8.3 error message

matthew-brett:
Message for binary file.

See: https://github.com/MacPython/pandas-wheels/issues/20#issuecomment-348809283


matthew-brett:
Merged with 155bb54

PR #30, opened 2018-01-11
First step toward windows support

xoviat:
This is the first step toward windows support. Let me know if you have any questions.


xoviat:
ping @matthew-brett

xoviat:
ping @matthew-brett

xoviat:
ping @matthew-brett

matthew-brett:
Sorry - just to check - only changes are, you've forced the dependency on setuptools, moved the scripts to be setuptools entrypoints, and added a couple of Windows-only dependencies?

xoviat:
That's correct.

> you've forced the dependency on setuptools

I think we're far enough along that depending on setuptools is not an issue.

matthew-brett:
Yes, setuptools is fine.  I don't personally like entrypoints very much, but it's just prejudice at this point (I tend to use scripts and https://matthew-brett.github.io/pydagogue/installing_scripts.html)

xoviat:
Entry points solve cross-platform fragility. Without them, you'd need to use hacks to get windows working.

matthew-brett:
I would say that entrypoints insert a rather weird hack for Windows, so the hack moves to setuptools, at the expense of transparency.  But I'll not fight you on it, I realize they are widely used.

anthrotype:
it's not just setuptools. pip also will create executable wrappers from console_script entry points, using a similar but different method than setuptools, which relies on [distlib](https://bitbucket.org/pypa/distlib)

matthew-brett:
As I say - it's only taste - the magicness of the system makes me uncomfortable - but I know that's not a source of suffering for everyone.

PR #34, opened 2018-01-31
Refactor tools and add test data

xoviat:


xoviat:
This is ready to merge.

xoviat:
cc @matthew-brett

xoviat:
cc @matthew-brett

delocate/tools/__init__.py
```diff
@@ -0,0 +1,44 @@
+import sys
+import os
+import zipfile
+
+from os.path import join as pjoin
+
+if sys.platform == 'win32':
```
matthew-brett: Not insisting, but it would be good to order these so it's obvious which names are specific to OSX / Windows, and which are not.

delocate/tools/win32.py
```diff
@@ -0,0 +1,183 @@
+import os
+import zipfile
+
+from bindepend import getImports
+from machomachomangler.pe import redll
+from .common import InstallNameError
+
+
+def get_install_names(filename):
+    """ Return install names from library named in `filename`
```
matthew-brett: Is 'install names' a thing in Windows?   Would it be better called 'DLL names'?   If not maybe an explanation in the docstring?

delocate/tools/win32.py
```diff
@@ -0,0 +1,183 @@
...
+    filename : str
+        filename of library
+    install_id : str
+        install id for library `filename`
+
+    Raises
+    ------
+    RuntimeError if `filename` has not install id
+    """
+    if get_install_id(filename) is None:
```
matthew-brett: Isn't it always `''` - from above?

delocate/tools/win32.py
```diff
@@ -0,0 +1,183 @@
...
+    """ Add rpath `newpath` to library `filename`
+
+    Parameters
+    ----------
+    filename : str
+        filename of library
+    newpath : str
+        rpath to add
+    """
+    return ''
```
matthew-brett: Why return the empty string?

delocate/tools/win32.py
```diff
@@ -0,0 +1,183 @@
...
+    -------
+    arch_names : frozenset
+        Empty (frozen)set if no arch codes.  If not empty, contains one or more
+        of 'ppc', 'ppc64', 'i386', 'x86_64'
+    """
+    # First, simple, implementation
+    return set('x86_64')
+
+
+def lipo_fuse(in_fname1, in_fname2, out_fname):
```
matthew-brett: This one doesn't make any sense on Windows ...

matthew-brett:
In general - I think this is a case where it would be more efficient for you to work on your own branch for a bit, to see how the API shakes out on Windows.  I'm very happy to have a look from time to time, if you think that would help.  I'll try to be more responsive these coming days.

xoviat:
> Is it really necessary to use the Mac names for these things? Or can the Windows low-levelimplementation use more Windows-like names, with the high-level API being common.

It's not so easy to develop a high-level API without any experience. Many of these functions are defined to avoid import errors (so that I can receive more meaningful errors during testing). This PR is mainly to get the diff down so that I can start working out the details.

matthew-brett:
I hear you about the API development - I have the same problem.   I bet that the combination of the tests and just getting it working will reveal where the Mac API sucks for Windows, and how we need to change it.

xoviat:
Do these changes significantly impair the OSX api? As I said, the point of this is mainly to reduce the diff, as my plain is to incrementally evolve the 'tools' api.

matthew-brett:
No, this doesn't change the OSX API, as you know, but my question is - is it worth merging this WIP, and then refactoring later, when the API becomes a bit clearer, or would it be better to leave this as a WIP, and keep working on it, merging when it's closer to its final shape?

Otherwise, delocate isn't changing very much, so I think you're not going to go stale very quickly.

xoviat:
Personally I'd be in favor of merging as WIP, as it makes the diff clearer when I do start refactoring.

matthew-brett:
Sorry - by WIP I mean a PR that is not yet to be merged.

How about we start a `devel` branch, and work on that instead, with PRs to `devel`?  Then we won't confuse anyone looking at the `master` branch.

xoviat:
Fine by me.

matthew-brett:
OK - I merged this branch into a new (version of the) `develop` branch ...  thanks.

PR #38, opened 2018-06-11
Process C extensions inside namespace packages

lpsinger:
When a wheel contains subpackages of a namespace package, there is no `__init__.py` file for the namespace package, though its subpackage directories do contain `__init__.py` files.

When scanning the wheel for package directories, searching the root for directories that contain `__init__.py` files is insufficient because we will miss C extension modules that are subpackages of namespace packages.

Instead, we need to do a depth-first traversal of the directory tree and capture the most shallow directories that contain `__init__.py` files.

### DIFF
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -404,10 +404,10 @@ def find_package_dirs(root_path):
         an ``__init__.py`` file.  Paths prefixed by `root_path`
     """
     package_sdirs = set()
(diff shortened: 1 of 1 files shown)


HexDecimal:
PR #123 was merged which has fixed this issue and includes tests for namespace packages.

lpsinger:
Nice! This worked for my package. Thank you!

PR #39, opened 2018-09-12
Handle C extensions not inside a package directory (at the root)

natefoo:
@matthew-brett how's this look for an implementation? It's working for me but needs a few tests.

Root packages end up with a `.dylib<pkg>` directory for delocated libs, this mirrors what auditwheel does.

### DIFF
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -3,11 +3,12 @@
 from subprocess import Popen, PIPE
 
 import os
-from os.path import join as pjoin, relpath, isdir, exists
+from os.path import join as pjoin, relpath, isdir, isfile, exists
 import zipfile
 import re
 import stat
 import time
+import warnings
 
 
 class InstallNameError(Exception):
@@ -399,15 +400,37 @@ def find_package_dirs(root_path):
 
     Returns
     -------
+    Set of strings where each is a subdirectory of `root_path`, containing
+    an ``__init__.py`` file.  Paths prefixed by `root_path`
+    """
+    warnings.warn("find_package_dirs() is deprecated, please use find_packages()", DeprecationWarning)
+    return [p for p, is_dir in find_packages(root_path) if is_dir]
+
+
+def find_packages(root_path, lib_filt_func=None):
+    """ Find python packages in directory `root_path`
+
+    Parameters
+    ----------
+    root_path : str
+        Directory to search for package subdirectories
+
+    Returns
+    -------
     package_sdirs : set
-        Set of strings where each is a subdirectory of `root_path`, containing
-        an ``__init__.py`` file.  Paths prefixed by `root_path`
+        Set of (string, is_dir) tuples where each string is a subdirectory of
+        `root_path` containing an ``__init__.py`` file, or a C extension
+        single-file module directly in `root_path`, and is_dir indicates whether
+        the package is a subdirectory module or a single-file module.  Paths
+        prefixed by `root_path`
     """
     package_sdirs = set()
     for entry in os.listdir(root_path):
         fname = entry if root_path == '.' else pjoin(root_path, entry)
         if isdir(fname) and exists(pjoin(fname, '__init__.py')):
-            package_sdirs.add(fname)
+            package_sdirs.add((fname, True))
+        elif isfile(fname) and dylibs_only(fname):
+            package_sdirs.add((fname, False))
     return package_sdirs
 
 
@@ -528,3 +551,8 @@ def validate_signature(filename):
 
     # This file's signature is invalid and needs to be replaced
     replace_signature(filename, '-') # Replace with an ad-hoc signature
+
+
+def dylibs_only(filename):
+    return (filename.endswith('.so') or
+            filename.endswith('.dylib'))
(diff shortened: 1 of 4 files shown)


delocate/tools.py
```diff
@@ -382,8 +382,8 @@ def dir2zip(in_dir, zip_fname):
     z.close()
 
 
-def find_package_dirs(root_path):
-    """ Find python package directories in directory `root_path`
+def find_packages(root_path):
```
matthew-brett: Sorry - I'm a little sensitive about changing the API without warning.

Would you mind making and deprecating something like:

```
def find_package_dirs(root_path):
    return [p for p, is_dir in find_packages(root) if is_dir]
```

and importing into `delocating`?

delocate/tools.py
```diff
@@ -393,14 +393,17 @@ def find_package_dirs(root_path):
     Returns
     -------
     package_sdirs : set
-        Set of strings where each is a subdirectory of `root_path`, containing
-        an ``__init__.py`` file.  Paths prefixed by `root_path`
+        Set of strings where each is a subdirectory of `root_path` containing an
```
matthew-brett: Actually a set of (string, is_dir) tuples no?

delocate/tools.py
```diff
@@ -393,14 +393,17 @@ def find_package_dirs(root_path):
...
+        ``__init__.py`` file, or a C extension single-file module directly in
+        `root_path`.  Paths prefixed by `root_path`
     """
     package_sdirs = set()
     for entry in os.listdir(root_path):
         fname = entry if root_path == '.' else pjoin(root_path, entry)
         if isdir(fname) and exists(pjoin(fname, '__init__.py')):
-            package_sdirs.add(fname)
+            package_sdirs.add((fname, True))
+        elif isfile(fname) and ((fname.endswith('.so') or fname.endswith('.dylib'))):
```
matthew-brett: Use `lib_filt_func` here?  See `delocate_wheel` docstring.

matthew-brett: I'm sure sure I understand, I might have been too hasty but `lib_filt_func` is to decide what files to consider when looking for libraries.  I guess it's fine print.   Yes, moving to `tools` sounds like a reasonable thing to do.

delocate/tools.py
```diff
@@ -393,14 +411,19 @@ def find_package_dirs(root_path):
...
+        the package is a subdirectory module or a single-file module.  Paths
+        prefixed by `root_path`
     """
     package_sdirs = set()
     for entry in os.listdir(root_path):
         fname = entry if root_path == '.' else pjoin(root_path, entry)
         if isdir(fname) and exists(pjoin(fname, '__init__.py')):
-            package_sdirs.add(fname)
+            package_sdirs.add((fname, True))
+        elif isfile(fname) and dylibs_only(fname):
```
matthew-brett: Here I was thinking you'd use lib_filt_func ...

natefoo: What should the behavior be if `lib_filt_func` is `None`, as is the default?

If I do:

```python
elif isfile(fname) and lib_filt_func and lib_filt_func(fname):
```

Then `delocate-wheel -d` is required in order to properly delocate top level C extension packages. Otherwise the behavior is the same as delocate currently works: top level C extension packages are included but not delocated.

If instead I do:

```python
elif isfile(fname) and ((lib_filt_func and lib_filt_func(fname)) or (lib_filt_func is None and dylibs_only(fname))):
```

Then it works without `-d`, but this seems against your intent as to the purpose of `lib_filt_func`?

Is the idea that `-d` should be required in order to delocate top level C extension packages? Or if `-d` is not used, should all files in the root be inspected, and in both cases, a more sophisticated method than `dylibs_only()` (i.e. binary inspection) should be used to determine if the file is a C extension package?

asottile:
looks like this has a conflict now -- I'm very interested in using this branch

matthew-brett:
@natefoo - OK to rebase? I promise to review quickly.

asottile:
the conflict is pretty trivial -- it's just mad about imports -- (can compare against https://github.com/asottile/delocate/tree/top-level-fix-squash_asottile)

natefoo:
@matthew-brett ping, I rebased this (also feel free to modify directly).

ipmb:
FWIW, I'm using this patch to build `pyuwsgi` wheels for Mac with success.

delocate/tools.py
matthew-brett: Yes, sorry, you're right, I forgot that `dylibs_only` is not the default.    I'll merge this manually, and do a few edits to the docstrings to remind myself what is going on.

natefoo:
@matthew-brett Anything I help with to get this over the line?

asottile:
reaching for this again -- unfortunately the conflict is pretty messy now so I'm having trouble rebasing this :(

bsolomon1124:
I'm working on rebasing this on a new branch, but https://github.com/matthew-brett/delocate/commit/9f042a87a8681dfeb1dde91143061f6ba010b7a3, which consisted of a refactor with `libsana.py`, is what's making that tedious.

McSinyx:
@bsolomon1124, thank you for continue working on this.  BTW recently per pypa/auditwheel#90, auditwheel has just adopted the new naming scheme for dynamic libraries (`{wheelname}.libs`).

natefoo:
Superseded by #123.

Issue #42, opened 2018-10-14
executable permissions lost after delocate-wheel


anthrotype:
I use multibuild (which uses delocate) to build a package that contains an embedded native executable, which is installed as package_data.

After delocate 0.7.4, the delocate-wheel command seems to make this file lose its executable permission flags, and thus when I attempt to call it as a subprocess I get PermissionError.

You can download this wheel to reproduce:

[psautohint-1.8.2a3-cp37-cp37m-macosx_10_13_x86_64.whl.zip](https://github.com/matthew-brett/delocate/files/2476325/psautohint-1.8.2a3-cp37-cp37m-macosx_10_13_x86_64.whl.zip)

Try pip installing this wheel in a fresh venv as it is, then try again after running `delocate-wheel` on it.
You'll notice that after the delocate-wheel command, the installed package contains a file `autohintexe` which does not have any executable flag set.

If I unzip the wheel, both before and after delocate-wheel command, I can see that the `autohintexe` command does have the executable permissions. It's only after running pip install on the delocated wheel, that the embedded executable loses the permission flags.

Thank you.

anthrotype:
Not sure if it's related or not but I noticed that, if I unzip the delocated wheel, the modified time is set to "1 Jan 1980" which is a bit strange. It doesn't happen with the wheel before running delocate-wheel.

```sh
$ ls -al psautohint
total 1096
drwxr-xr-x  10 cosimolupo  staff     320 14 Oct 12:46 ./
drwxr-xr-x   6 cosimolupo  staff     192 14 Oct 12:46 ../
-rw-r--r--   1 cosimolupo  staff    4825  1 Jan  1980 __init__.py
-rw-r--r--   1 cosimolupo  staff   31147  1 Jan  1980 __main__.py
-rwxr-xr-x   1 cosimolupo  staff  191336  1 Jan  1980 _psautohint.cpython-37m-darwin.so*
-rw-r--r--   1 cosimolupo  staff   22646  1 Jan  1980 autohint.py
-rwxr-xr-x   1 cosimolupo  staff  194432  1 Jan  1980 autohintexe*
-rw-r--r--   1 cosimolupo  staff   19984  1 Jan  1980 fdTools.py
-rw-r--r--   1 cosimolupo  staff   32971  1 Jan  1980 otfFont.py
-rw-r--r--   1 cosimolupo  staff   47227  1 Jan  1980 ufoFont.py
```

anthrotype:
[...]

PR #43, opened 2018-10-14
set S_IFREG (regular file) in ZipInfo.external_attr

anthrotype:
Fixes https://github.com/matthew-brett/delocate/issues/42

When zipping the wheel, we need to set `stat.S_IFREG` (for regular files) in the `ZipInfo.external_attr` attribute.
If we don't do that, then upon installing the wheel with pip, any files that had executable flags (e.g. embedded executables) in the wheel will lose them.
Pip restores the execute permissions when uncompressing the wheel only if `stat.S_ISREG(mode)` is True:

https://github.com/pypa/pip/blob/35f098363983baebc13ffb84ded80b816d8fb246/src/pip/_internal/utils/misc.py#L488-L493

### DIFF
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -375,7 +375,7 @@ def dir2zip(in_dir, zip_fname):
             info = zipfile.ZipInfo(in_fname)
             info.filename = relpath(in_fname, in_dir)
             # See https://stackoverflow.com/questions/434641/how-do-i-set-permissions-attributes-on-a-file-in-a-zip-file-using-pythons-zip/48435482#48435482
-            info.external_attr = chmod_perms(in_fname) << 16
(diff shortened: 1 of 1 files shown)


matthew-brett:
Thanks for tracking that down.

JonasVautherin:
This patch never got released, did it? Seems like the last release is from Oct 1, 2018 =/.

matthew-brett:
Right - yes - I haven't done a delocate release for a while.  I have some marking to finish up, and then I'll be spending some time clearing out my open-source queue - I hope at the end of this week.

JonasVautherin:
Ok thanks! Building from source in the meantime, but would be awesome to have the latest version on pip :+1:

JonasVautherin:
Still thinking it may make sense to release that :see_no_evil:.

PR #44, opened 2018-10-15
WIP: tests for file type / permissions in wheel

matthew-brett:
Tests for #43.  But - even a original wheel build doesn't seem to keep the
executable permissions for a script inside the package directory.

@anthrotype - any ideas what's going on here?  Do these look like the right
tests?


anthrotype:
I'll take a look. Why isn't Travis CI enabled on this PR?

anthrotype:
I see. the `fakepkg1/ascript.py` is losing its executable permissions when making the original wheel file (before delocating).
In my project, I'm actually building a native c executable, not a plain text python script. Maybe that's the difference. I suggest you try to compile a minimalist hello world C executable and bundle that in the test fakepkg1.

anthrotype:
But I can tell you that the #43 patch works for our use case.
This is the project in case you want to have a look:
https://github.com/adobe-type-tools/psautohint

matthew-brett:
What command do you use to build your wheel file?

anthrotype:
it's using multibuild's default command, which I believe is `pip wheel --no-deps .`

matthew-brett:
@anthrotype - would you mind having a look at the current state?  I'm also setting and checking the modification times.

PR #46, opened 2018-10-28
remove wheeltools.py, import/require standalone wheeltools

anthrotype:
as discussed in https://github.com/pypa/auditwheel/issues/102

I published a pre-release version to PyPI so that we can try this out in delocate:
https://pypi.org/project/wheeltools/


delocate/tools.py
```diff
@@ -1,132 +1,25 @@
...
+from wheeltools.tools import (
+    back_tick,
+    ensure_writable,
+    find_package_dirs,
+    dir2zip,
+    zip2dir,
+    cmp_contents,
+    open_readable,
+    open_rw,
+    chmod_perms,
```
anthrotype: @matthew-brett I kept these names in here as aliases for the functions that were moved out to the standalone wheeltools package. If you prefer, we can instead fix up all the relative `.tools` imports in delocate to instead import these from the `wheeltools.tools`.

setup.py
```diff
@@ -31,6 +31,7 @@
           "machomachomangler; sys_platform == 'win32'",
           "bindepend; sys_platform == 'win32'",
           "wheel",
+          "wheeltools >= 0.1.0b1",
```
anthrotype: we can just write `wheeltools` in install_requires, once we publish a stable release

anthrotype:
Judging from the activity on auditwheel repo, and the lack of interest/feedback its maintainers showed for this splitting up and reuse of delocate’s wheeltools module in auditwheel, i think it’s not worth to proceed with this. Feel free to close this PR.

matthew-brett:
Yes, sorry - I think the work involved is practical, but I can't do it at the moment, I am completely buried in work.

For @ehashman - I still think this is the right way to go,and would add tests to a part of auditwheel that doesn't have them.

anthrotype:
I understand, no problem!

matthew-brett:
I guess we could try again on the Auditwheel list to see if we can get more joint traction.

PR #54, opened 2019-07-19
Updated Python versions in README

radarhere:


matthew-brett:
Thanks for the reminder here - I've update the Python versions (and travis file) in current master, so closing here.

Issue #49, opened 2019-01-23
delocate-wheel can't add dylibs and so files


polychordlondon:
I'm trying to build a wheel for MacOS users, the source code of this wheel can be found here:
https://github.com/PolyChord/PolyChordLite/tree/setup_pip

I'm facing the same problem than [issue#15](https://github.com/matthew-brett/delocate/issues/15):

    ➜  polychordlite git:(setup_pip) delocate-listdeps --all --depending dist/pypolychord_nompi-1.15.1-cp27-cp27m-macosx_10_12_x86_64.whl                                                                        
    /Users/polychord/Desktop/polychordlite/pypolychord/lib/libchord.so:
        _pypolychord.so
    /usr/lib/libSystem.B.dylib:
        _pypolychord.so
    /usr/local/Cellar/gcc/8.2.0/lib/gcc/8/libgcc_s.1.dylib:
        _pypolychord.so
    /usr/local/Cellar/gcc/8.2.0/lib/gcc/8/libgfortran.5.dylib:
        _pypolychord.so
    /usr/local/Cellar/gcc/8.2.0/lib/gcc/8/libstdc++.6.dylib:
        _pypolychord.so

    ➜  polychordlite git:(setup_pip) delocate-wheel  -w dist/fixed -v dist/pypolychord_nompi-1.15.1-cp27-cp27m-macosx_10_12_x86_64.whl                                                                           
    Fixing: dist/pypolychord_nompi-1.15.1-cp27-cp27m-macosx_10_12_x86_64.whl

Nothing is added to the fixed wheel.

But I think that the architecture is fine (architecture of the wheel):

    _pypolychord.so
    pypolychord
        __init__.py
        output.py
        priors.py
        settings.py
    pypolychord_nompi-1.15.1.dist-info
        ...

The wheels can be built on linux with auditwheel.

Python 2.7 - delocate 0.7.4 - MacOS 10.12.6 (VM)

zhreshold:
+1, same behavior here, with `-v` it shows `Fixing: dist/xxx.whl` but nothing happens

PR #55, opened 2019-09-12
fix with recursive search

zhreshold:
Solve problem here: https://github.com/matthew-brett/delocate/issues/49

The root cause is that `find_package_dirs` only list the level-1 dir of the wheel which in some cases are not sufficient. There are times `library` in inside `purelib` or `platlib`

### DIFF
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -408,10 +408,10 @@ def find_package_dirs(root_path):
         an ``__init__.py`` file.  Paths prefixed by `root_path`
     """
     package_sdirs = set()
-    for entry in os.listdir(root_path):
-        fname = entry if root_path == '.' else pjoin(root_path, entry)
-        if isdir(fname) and exists(pjoin(fname, '__init__.py')):
-            package_sdirs.add(fname)
+    # use recursive search in case there's purelib or platlib layers
+    for entry in [dp for dp, _, _ in os.walk(os.path.expanduser(root_path))]:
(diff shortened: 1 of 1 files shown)


matthew-brett:
Thanks for the fix.  You probably saw - but there are a lot of test failures - could you look at those?

Do you have a wheel that fails?   And that you can add as a test?

The ideal would be to build the wheel for the test, so we can keep it up to date.  See the `wheel_makers` directory.

jnak:
FYI I tried your branch and it did not fix my issue (same as https://github.com/matthew-brett/delocate/issues/49). The library I'm trying to build is PyICU:
```
➜  wheels  delocate-listdeps PyICU-2.3.1-cp27-cp27m-macosx_10_12_x86_64.whl
/usr/local/Cellar/icu4c/64.2/lib/libicudata.64.2.dylib
/usr/local/Cellar/icu4c/64.2/lib/libicui18n.64.2.dylib
/usr/local/Cellar/icu4c/64.2/lib/libicuuc.64.2.dylib
➜  wheels delocate-wheel -w fixed_wheels -v PyICU-2.3.1-cp27-cp27m-macosx_10_12_x86_64.whl
Fixing: PyICU-2.3.1-cp27-cp27m-macosx_10_12_x86_64.whl
```

zhreshold:
Let me figure out a way to fit broader tests, so far it only works with my wheel. I will reopen if I find other solutions.

PR #58, opened 2019-10-07
Update "not an object" table for cctools-949

atwilso:
Sometime between cctools-900 and cctools-949, the `otool` utility on
OS X made a one-character change in its 'this isn't an object file'
message.  The old message was:

'Object is not a Mach-O file type'

and the new one is

'object is not a Mach-O file type'.

I've added a new line to `BAD_OBJECT_TESTS` in tools.py to check for
the new version.


matthew-brett:
Thanks for working that out.

PR #60, opened 2019-11-20
fix PEP 8  

Czaki:
In this PR I fix mos of errors reported by flake8 tool and mark others to not be reported (with `.flake` file and `noqa` directives


delocate/tmpdirs.py
```diff
@@ -1,13 +1,14 @@
 # emacs: -*- mode: python-mode; py-indent-offset: 4; indent-tabs-mode: nil -*-
 # vi: set ft=python sts=4 ts=4 sw=4 et:
+"""
```
matthew-brett: I don' t think we do want this stuff in the docstring.

matthew-brett:
Thanks - in it goes.

PR #61, opened 2019-12-14
Fix weel name using otool

Czaki:
Fixing wheel name with parsing header from python code (based on pypa/wheel#314) and calculate proper platform tag.

 - [x] parsing headers to extract minimal version
 - [x] test of added functions 
 - [ ] new test for `delocate_wheel`


Czaki:
`otool` from old xcode does not support `LC_BUILD_VERSION`

HexDecimal:
Resolved by #198

Issue #15, opened 2017-02-22
delocate-wheel do nothing on wheel


guyskk:
I'm building macosx wheels for lightpipes, the source code is:
https://github.com/guyskk/lightpipes/tree/wheels_for_macos

lightpipes need fftw, I install fftw by `brew install fftw`.
to build the wheel, run `pip install numpy` then `pip wheel --no-deps -w wheelhouse ./lightpipes` .
after that, I use delocate to fix the wheel:

```bash
➜  lightpipes git:(wheels_for_macos) delocate-listdeps --all  --depending wheelhouse/LightPipes-1.1.0-cp27-cp27m-macosx_10_10_x86_64.whl
/usr/lib/libSystem.B.dylib:
    LightPipes.so
/usr/lib/libc++.1.dylib:
    LightPipes.so
/usr/local/Cellar/fftw/3.3.6-pl1/lib/libfftw3.3.dylib:
    LightPipes.so
➜  lightpipes git:(wheels_for_macos) delocate-wheel -w wheelhouse_fixed -v wheelhouse/LightPipes-1.1.0-cp27-cp27m-macosx_10_10_x86_64.whl
Fixing: wheelhouse/LightPipes-1.1.0-cp27-cp27m-macosx_10_10_x86_64.whl
```

but it do nothing on the wheel:
```bash
➜  lightpipes git:(wheels_for_macos) ls wheelhouse_fixed
LightPipes-1.1.0-cp27-cp27m-macosx_10_10_x86_64.whl
➜  lightpipes git:(wheels_for_macos) ls -lh wheelhouse_fixed
total 176
-rw-r--r--  1 xiachufang  staff    86K  2 22 17:53 LightPipes-1.1.0-cp27-cp27m-macosx_10_10_x86_64.whl
➜  lightpipes git:(wheels_for_macos) ls -lh wheelhouse
total 176
-rw-r--r--  1 xiachufang  staff    86K  2 22 17:53 LightPipes-1.1.0-cp27-cp27m-macosx_10_10_x86_64.whl
```

matthew-brett:
I think the problem is that your package has a single module at the top level (that goes in the `site-packages` directory, and delocate doesn't deal with this case - see https://github.com/matthew-brett/delocate/issues/12#issuecomment-246536098

To get delocate to work, you could reorganize your package like this:

```
setup.py
lightpipes/
    __init__.py
    LightPipes.pyx
    etc
```

[...]

Issue #12, opened 2016-09-12
delocate-wheel and non-fixed brew dependency


hredestig:
I am trying to use multibuild (great project btw, thanks a lot for this one) to make wheels for [swiglpk](https://github.com/biosustain/swiglpk) but have difficulty to fix the library dependencies for the mac wheels. A reproducible example for is

``` sh
$ git clone https://github.com/biosustain/swiglpk
$ brew install glpk
$ cd swiglpk/
$ python setup.py bdist_wheel
$ delocate-listdeps dist/swiglpk-1.2.22+4.g954dc63-cp35-cp35m-macosx_10_6_intel.whl 
/usr/local/Cellar/glpk/4.60/lib/libglpk.40.dylib
$ delocate-wheel -v -w fixed-wheel dist/swiglpk-1.2.22+4.g954dc63-cp35-cp35m-macosx_10_6_intel
$ delocate-listdeps fixed-wheel/swiglpk-1.2.22+4.g954dc63-cp35-cp35m-macosx_10_6_intel.whl 
/usr/local/Cellar/glpk/4.60/lib/libglpk.40.dylib
```

With the last call to `delocate-listdeps` I expected no output indicating that `delocate-wheel` had copied the dependency to the wheel. Am I missing something or is this a bug somehow?

matthew-brett:
Does the wheel in fact contain the library?  I seem to remember that delocate does not by default print out the libraries it copied, perhaps I should change that line in multibuild to avoid the confusion.

hredestig:
no, pardon the confusion, the fixed wheel doesn't contain the library so something is going wrong when I do `delocate-wheel`

matthew-brett:
Pardon me for not reading your message with enough attention.  I'll have a look later on today.

matthew-brett:
Ah - yes - I see.  The wheel processor specifically looks in the subdirectories within the wheel, and not in the compiled modules at the top level.  The reason is that it wants to create a new subdirectory, called `.libs` with the copied libraries in.  This is unique to the package if it can work in a subdirectory (`.../site-packages/yourpkg/.libs/copied_lib.dylib`), but may well not be unique for a module at the top level (`.../site-packages/.libs/copied_lib.dylib`).  So, to make it easier for delocate, you could make the module into its own package.  Otherwise, I guess I'd need to work out some naming scheme for the generated lib / lib directory.

hredestig:
Alright, thanks that makes sense, will try to adjust on our end

[...]

PR #70, opened 2020-04-09
Doc limitation with top-level Py module structure

bsolomon1124:
As discussed in Issue #12, Issue #15, and elsewhere
delocate-wheel may ignore compiled extensions at the
top level of a wheel that are not part of a Python
package.  This documents that limitation and shows
how the package structure can be adjuste to fix it.


matthew-brett:
Nice - thanks!

PR #76, opened 2020-05-31
add escape of character in name of library

Czaki:
Fix regexp to escape characters for names like `name++`


matthew-brett:
Sorry about the delay here - looks good.

PR #91, opened 2020-12-31
Raise an exception on dependency resolution failures.

### LINKED ISSUE #90: delocate-wheel exits 'succesfully' after encountering potential errors.
I've recently ran into a problem where updating a dependency I was using added another dependency  which failed to be included by `delocate-wheel`.  Because `delocate-wheel` doesn't exit with an error code on warnings I didn't find out there was an issue until sometime later.  Example output:
```
delocate-wheel -v dist/*.whl
/Users/runner/work/python-tcod/python-tcod/venv/lib/python3.7/site-packages/delocate/libsana.py:111: UserWarning: Couldn't find @rpath/hidapi.framework/Versions/A/hidapi on paths:
Fixing: dist/tcod-11.19.1-cp35-abi3-macosx_10_9_x86_64.whl
	/Users/runner/work/python-tcod/python-tcod/@executable_path/Frameworks
Copied to package .dylibs directory:
	/Users/runner/work/python-tcod/python-tcod/@loader_path/Frameworks
  '\n\t'.join(realpath(path) for path in rpaths),
  /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
/Users/runner/work/python-tcod/python-tcod/venv/lib/python3.7/site-packages/delocate/libsana.py:111: UserWarning: Couldn't find @rpath/hidapi.framework/Versions/A/hidapi on paths:
/System/Library/Frameworks/AppKit.framework/Versions/C/AppKit
	/Users/runner/work/python-tcod/python-tcod/@executable_path/Frameworks
/System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices
	/Users/runner/work/python-tcod/python-tcod/@loader_path/Frameworks
/System/Library/Frameworks/AudioToolbox.framework/Versions/A/AudioToolbox
  '\n\t'.join(realpath(path) for path in rpaths),
```
I'd like to suggest a command line argument to determine if warnings like these return an error code instead of passing silently in scripts, or maybe even change the default behavior of when libraries aren't found to return an error by default with an option to allow the behavior.

I think I could implement this myself if you decide what the policy should be.

HexDecimal:
This only affects libraries with `@rpath` dependencies.  It's also me fixing my own bad code when I wrote resolve_rpath.  At least I knew this could be a problem back then and added a warning for it.

Previously this issue was warned about using Python's warning module, but these errors are more often critical and will affect the compatibility of the wheel.  A failure here means that either the library actually is missing, delocate is not being correct, or the library was built incorrectly.

The more I considered it the less I was able to justify the ability to ignore these errors as an option since it's too implicit and could easily ignore unintended dependencies.  I did add a parameter to tree_libs which can be enabled to ignore these errors, but there's no new command line options yet.

This is what the new error looks like in practice, compared to the example in #90:
```
delocate-wheel -v dist/*.whl
Fixing: dist/tcod-11.19.2.dev2-cp35-abi3-macosx_10_9_x86_64.whl
  File "/Users/runner/work/python-tcod/python-tcod/venv/bin/delocate-wheel", line 8, in <module>
    sys.exit(main())
  File "/Users/runner/work/python-tcod/python-tcod/venv/lib/python3.7/site-packages/delocate/cmd/delocate_wheel.py", line 72, in main
    check_verbose=opts.verbose)
  File "/Users/runner/work/python-tcod/python-tcod/venv/lib/python3.7/site-packages/delocate/delocating.py", line 378, in delocate_wheel
    lib_filt_func, copy_filt_func)
  File "/Users/runner/work/python-tcod/python-tcod/venv/lib/python3.7/site-packages/delocate/delocating.py", line 290, in delocate_path
    return copy_recurse(lib_path, copy_filt_func, copied)
  File "/Users/runner/work/python-tcod/python-tcod/venv/lib/python3.7/site-packages/delocate/delocating.py", line 146, in copy_recurse
    _copy_required(lib_path, copy_filt_func, copied_libs)
  File "/Users/runner/work/python-tcod/python-tcod/venv/lib/python3.7/site-packages/delocate/delocating.py", line 202, in _copy_required
    lib_dict = tree_libs(lib_path)
  File "/Users/runner/work/python-tcod/python-tcod/venv/lib/python3.7/site-packages/delocate/libsana.py", line 103, in tree_libs
    raise DependencyNotFound(error_msg)
delocate.libsana.DependencyNotFound: Could not find these dependencies:
@rpath/hidapi.framework/Versions/A/hidapi not found:
  Needed by: /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpabvk0pzn/wheel/tcod/.dylibs/SDL2
  Search path:
    @executable_path/Frameworks
    @loader_path/Frameworks
```
This error reveals what happened.  SDL2 now has a Frameworks directory embedded in its own framework and delocate doesn't support `@loader_path` yet.  This will need to be fixed so I can update SDL to the latest version, but I'm not a hurry.  Anyone else might be able to use `DYLD_LIBRARY_PATH` as a workaround.

As mentioned before this could possibly be a breaking change.  I've given my reasons for why this is acceptable but tree_libs can be told to ignore errors if you can pass the option down to it.

Resolves #90

### DIFF
--- a/delocate/libsana.py
+++ b/delocate/libsana.py
@@ -6,15 +6,22 @@ Analyze library dependencies in paths and wheel files
 import os
 from os.path import basename, join as pjoin, realpath
 
-import warnings
-
 from .tools import (get_install_names, zip2dir, get_rpaths,
                     get_environment_variable_paths)
 from .tmpdirs import TemporaryDirectory
 
 
-def tree_libs(start_path, filt_func=None):
+class DependencyNotFound(Exception):
+    """Raised by tree_libs or resolve_rpath if an expected dependency is
+    missing.
+    """
+
+
+def tree_libs(start_path, filt_func=None, ignore_missing=False):
     """ Return analysis of library dependencies within `start_path`
+    
+    DependencyNotFound is raised if any `@rpath` dependencies can not be
+    resolved.
 
     Parameters
     ----------
@@ -24,6 +31,9 @@ def tree_libs(start_path, filt_func=None):
         If None, inspect all files for library dependencies. If callable,
         accepts filename as argument, returns True if we should inspect the
         file, False otherwise.
+    ignore_missing : bool
+        Determines if dependency resolution failures are considered a critical
+        error to be raised.  Otherwise ignored libraries are only printed out.
 
     Returns
     -------
@@ -50,6 +60,7 @@ def tree_libs(start_path, filt_func=None):
     * http://matthew-brett.github.io/pydagogue/mac_runtime_link.html
     """
     lib_dict = {}
+    missing_dependencies = []  # Info on libraries not found by resolve_rpath.
     env_var_paths = get_environment_variable_paths()
     for dirpath, dirnames, basenames in os.walk(start_path):
         for base in basenames:
@@ -63,7 +74,13 @@ def tree_libs(start_path, filt_func=None):
                 # We'll do nothing to other '@'-paths
                 # Otherwise we'll search for the library using env variables
                 if install_name.startswith('@rpath'):
-                    lib_path = resolve_rpath(install_name, search_paths)
+                    try:
+                        lib_path = resolve_rpath(install_name, search_paths)
+                    except DependencyNotFound:
+                        lib_path = install_name
+                        missing_dependencies.append(
+                            (install_name, search_paths, depending_libpath)
+                        )
                 elif install_name.startswith('@'):
                     lib_path = install_name
                 else:
@@ -72,13 +89,29 @@ def tree_libs(start_path, filt_func=None):
                     lib_dict[lib_path][depending_libpath] = install_name
                 else:
                     lib_dict[lib_path] = {depending_libpath: install_name}
+
+    if not ignore_missing and missing_dependencies:
+        error_msg = "Could not find these dependencies:"
+        for missing, rpaths, dependant in missing_dependencies:
+            error_msg += (
+                "\n{0} not found:"
+                "\n  Needed by: {1}"
+                "\n  Search path:\n    {2}".format(
+                    missing, dependant, "\n    ".join(rpaths)
+                )
+            )
+        raise DependencyNotFound(error_msg)
+
+    for missing, _, _ in missing_dependencies:
+        print("Ignoring missing dependency: {0}".format(missing))
+
     return lib_dict
 
 
 def resolve_rpath(lib_path, rpaths):
     """ Return `lib_path` with its `@rpath` resolved
 
-    If the `lib_path` doesn't have `@rpath` then it's returned as is.
+    If the `lib_path` doesn't have `@rpath` then DependencyNotFound is raised.
 
     If `lib_path` has `@rpath` then returns the first `rpaths`/`lib_path`
     combination found.  If the library can't be found in `rpaths` then a
@@ -105,13 +138,7 @@ def resolve_rpath(lib_path, rpaths):
         if os.path.exists(rpath_lib):
             return rpath_lib
 
-    warnings.warn(
-        "Couldn't find {0} on paths:\n\t{1}".format(
-            lib_path,
-            '\n\t'.join(realpath(path) for path in rpaths),
-            )
-        )
-    return lib_path
+    raise DependencyNotFound(lib_path)
 
 
 def search_environment_for_lib(lib_path):
--- a/delocate/libsana.py
+++ b/delocate/libsana.py
@@ -4,7 +4,7 @@ Analyze library dependencies in paths and wheel files
 """
 
 import os
-from os.path import basename, join as pjoin, realpath
+from os.path import basename, dirname, join as pjoin, realpath
 
 import warnings
 from typing import Callable, Dict, Iterable, List, Optional, Text
@@ -93,7 +93,11 @@ def tree_libs(
                 # Otherwise we'll search for the library using env variables
                 if install_name.startswith('@rpath'):
                     try:
-                        lib_path = resolve_rpath(install_name, search_paths)
+                        lib_path = resolve_rpath(
+                            install_name,
+                            search_paths,
+                            loader_path=dirname(depending_libpath),
+                        )
                     except DependencyNotFound:
                         lib_path = install_name
                         missing_dependencies.append(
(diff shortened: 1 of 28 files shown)


delocate/libsana.py
```diff
@@ -24,6 +31,9 @@ def tree_libs(start_path, filt_func=None):
         If None, inspect all files for library dependencies. If callable,
         accepts filename as argument, returns True if we should inspect the
         file, False otherwise.
+    ignore_missing : bool
```
matthew-brett, line 34: This should be:

```
ignore_missing : {False, True}, optional
```

matthew-brett, line 34: Do we want to ignore or warn?

delocate/libsana.py
```diff
@@ -24,6 +31,9 @@ def tree_libs(start_path, filt_func=None):
...
+        Determines if dependency resolution failures are considered a critical
+        error to be raised.  Otherwise ignored libraries are only printed out.
```
matthew-brett, line 36: Could you also add a `Raises` section at the end for the `DependencyNotFound` error?  <https://numpydoc.readthedocs.io/en/latest/format.html>

delocate/tests/test_libsana.py
```diff
@@ -182,5 +183,5 @@ def test_resolve_rpath():
     lib_rpath = pjoin('@rpath', lib)
     # Should skip '/nonexist' path
     assert_equal(resolve_rpath(lib_rpath, ['/nonexist', path]), realpath(LIBA))
-    # Should return the given parameter as is since it can't be found
-    assert_equal(resolve_rpath(lib_rpath, []), lib_rpath)
+    # Should raise DependencyNotFound if the dependency can not be resolved.
```
matthew-brett, line 186: And test for warning, if that seems useful?

matthew-brett (commented):
Thanks for this - I agree, it seems better to raise here, despite back-compatibility.

delocate/libsana.py
HexDecimal: Sorry for that.  Right now I'm more used to MyPy type-hinting where it's normally bool, and optional has a different meaning.

HexDecimal: Currently `tree_libs` prints out ignored dependencies.  I don't want it to be silent since it seems very important when they get ignored.  Maybe 'skip_missing' would be a better term?

delocate/libsana.py
HexDecimal: I almost added a raises section but I didn't find anything else in the module to reference when I looked and didn't notice this was using the NumPy docstring style guide.  I'll add one now.

A contributing file might have helped.

delocate/tests/test_libsana.py
HexDecimal: The skip missing behavior is in the `tree_libs` function.  I changed the feedback from using the warnings module to just print, since it'd be expected behavior by that point.

HexDecimal: I have it back to using warnings now.

HexDecimal:
I'm thinking of working on the `@loader_path` issue in another PR.  Would it be okay if I setup GitHub Actions for Flake8 and MyPy?  I'll need MyPy so that I can more easily work on larger refactors.

matthew-brett:
Yes, please go ahead ... thanks.

delocate/tests/test_libsana.py
ferdnyc: @HexDecimal 
I checked out this branch to compare the PR code with the released `0.8.2`, after hitting this output when running `delocate-listdeps`. I have to say, it doesn't feel particularly right to me, in either the previous or current form. (Though it's definitely _much_ improved, now that it's no longer outputting module source code as the warning message!)

Even with the fixed code, though, it doesn't feel like the right sort of output _when running an interactive utility_. (Meaning, in the context of `delocate-listdeps` specifically.) In that context, in fact, it feels a bit like a violation of the ["When to use logging"](https://docs.python.org/3/howto/logging.html#when-to-use-logging) chart from the Basic logging tutorial, which advises:

|Task you want to perform                       | The best tool for the task  |
|------------------------------------------------|----------------------------------|
| Display console output for ordinary usage of a command line script or program | `print()` |
| Issue a warning regarding a particular runtime event | `warnings.warn()` in library code if the issue is avoidable and the client application should be modified to eliminate the warning |

...The primary implication here being that the warning signals an _application_ error, that "the client application should be modified to eliminate the warning". There's no change that can be made to `delocate-listdeps` to correct unresolvable dependency paths in a native binary. Shouldn't that just be reported in the normal `print()` output, like the rest of the dependency listing?

HexDecimal: I've been having the same feeling and I wish the logging module itself was being used.  There's a lot of stuff I wished I could print for debugging without it being in the main output.

Would it be valid to import the logging module and start using it for this kind of thing?  Or should I just stick with print?

matthew-brett: It's a bit tricky, because of the usual tension between libraries and command line utilities.   Printing from libraries is usually very bad style, and delocate can be used as a library.   So, what I'd like to find is some way of making the output behave as expected for a command line utilty - with ordinary, expected output going to stdout, and warnings / errors going to stderr - while also making it behave sensibly when used as a library.

ferdnyc: I think, precisely because of that multi-purpose utility, **no** decision about how to respond to a missing dependency can really be made at that low a level -- not to the point of informing the user, at least. Not by any method. 

Because, if the caller id `delocate-listdeps`, then an unresolvable path is merely a point of information, no more or less bad than a resolved path, and the tool should handle informing the user of both through normal means. There's no error of any kind, in that situation.

But if the caller is `delocate-wheel` trying to fix up package's external dependencies, not being able to resolve one is a four-alarm failure condition. So **then** it needs to be shouted about, in a way that user can't possibly miss.

That's a decision that can only be made higher up the call chain, with benefit of the proper context to make the right call. **Any** action taken way down in the inner workings of the library _other_ than kicking the problem upstairs will invariably be the wrong one, for some subset of all the possible scenarios.

matthew-brett:
But - isn't that what try : catch is for?

ferdnyc:
@matthew-brett 

That's a fair question. ...Maybe?

Problem with raising is, it breaks you out of the current operation. That's fine if the current operation is "report the status of this individual dependency", but certainly nobody'd want `delocate-listdeps` to give up at the first sign of an unresolved dep.

And the other side of that coin is, if a dependency resolution failure is going to raise an exception, then it needs to be raised immediately, each time there's a failure, but without preventing the rest of the lookups from completing.

HexDecimal:
I had issues with raising errors in practice and had to stop.  PR #94 continues this branch, but with the topic of `@loader_path` support instead.

Issue #7, opened 2014-10-12
Recursive dependencies not shown in delocate-listdeps


letmaik:
I'm writing a Python extension for a native library. delocate-listdeps lists the native library, but it doesn't list the dependencies of the native libary which would have to be included as well in the wheel. How should I do that? Is that supported already?

matthew-brett:
Sorry - I should also say that ``delocate-listdeps --all`` will do what you want.

tschiex:
I have a closely related strange behavior in delocate. My python package includes a (pybind11 glued) library that depends on libboost-graph that depends on libicu (including libicui18n.64.2.dylib).  libboost-graph and libicui18n.64.2.dylib are correctly added to the .dylib folder, but libicui18n.64.2.dylib depends on libicudata.64.dylib and libicuuc.64.dylib and they are not included in the .dylib folder. Unsurprisingly, importing the python package leads to an import error, indicating the @loader_path/libicuuc.64.dylib, referenced from  libicui18n.64.2.dylib, cannot be loaded.
Is there a limit to the recursion of delocate?

matthew-brett:
I don't know of a limit.  It sounds like we need to investigate ...

tschiex:
Actually, after reading delocating and libsana, I think the issue comes from the fact that delocate does not handle @loader_path dependencies. In my case,  libicui18n.64.2.dylib depends on @loader_path/libicudata.64.dylib and @loader_path/ libicuuc.64.dylib. These are not copied (or renamed). Indeed, their realnames are actually /libicudata.64.1.dylib and  libicuuc.64.1.dylib which have been copied (because of other dependecies).

I would say that when a library is copied that has original @loader_path dependencies, if these dependencies' realnames have been copied, then the install name they are referred by should be changed to the new local .dylib name. Otherwise, the library should be copied too.

tschiex:
Actually, it looks like[ HexDecimal PR](https://github.com/matthew-brett/delocate/pull/94) above could solve my issue.

tschiex:
Well, just tried and.. no. It doesn't.  The copied libicui18n.64.2.dylib still depends on @loader_path/libicudata.64.dylib and @loader_path/ libicuuc.64.dylib which are copied under their realnames, not this  @loader_path that refers to a symbolic link.

mathstuf:
[...]

PR #94, opened 2021-01-11
Support for `@loader_path`.

### LINKED ISSUE #37: delocating wheels built with conda's clang
Consider the following example:

foo.cpp (trivial module that links to libc++):
```
#include "Python.h"
#include <iostream>

static struct PyModuleDef moduledef = {
  PyModuleDef_HEAD_INIT,
  "foo", NULL, 0, NULL, NULL, NULL, NULL, NULL
};

PyMODINIT_FUNC PyInit_foo() {
  std::cout << "hello world\n";  // enforce that we'll link to libc++
  return PyModule_Create(&moduledef);
}
```

setup.py:
```
from setuptools import *
setup(name="foo",
      packages=["pkg"],
      ext_modules=[Extension("pkg.foo", ["foo.cpp"],
                             extra_compile_args=[
                                 "-mmacosx-version-min=10.9",
                                 "-stdlib=libc++"],
                             extra_link_args=["-lc++"])])
```

I explicitly link against libc++ because we want it to get packaged by delocate-wheel.  The reason is that I actually rely on a very recent libc++ (for C++17 support in https://github.com/anntzer/mplcairo).  The equivalent for a Linux build is to pass `-static-libgcc -static-libstdc++` to gcc, but AFAICT libc++ cannot be statically linked.

Let's try to build this using conda's clang (I don't have the permission to use homebrew on that machine) in a clean environment:
```
conda create -yn tmp python=3.6 &&
    activate tmp &&
    conda install -yc anaconda clangxx_osx-64 &&
    pip install delocate
```
then `python setup.py bdist_wheel` and `delocate-wheel dist/foo-*.whl`.

This fails with
```
/Users/anntzer/miniconda3/envs/mplcairo-build/lib/python3.6/site-packages/delocate/libsana.py:101: UserWarning: Couldn't find @rpath/libc++abi.1.dylib on paths:
        /Users/anntzer/src/local/miniextension/@loader_path
  '\n\t'.join(realpath(path) for path in rpaths),
Copied to package .dylibs directory:
  /Users/anntzer/miniconda3/envs/tmp/lib/libc++.1.0.dylib
```

Note, interestingly, that if the extension module is at the toplevel (not placed in a `pkg` package), then delocate-wheel does not complain.

### LINKED ISSUE #62: delocate fails when LC_RPATH is set to @loader_path
Using a lib from [ortools](https://github.com/google/or-tools/releases/download/v7.4/or-tools_MacOsX-10.14.6_v7.4.7247.tar.gz)
```
$ otool -l libClp.1.14.3.dylib
...
          cmd LC_RPATH
      cmdsize 32
         path @loader_path (offset 12)
```
`delocate-wheel` is confused, searching for a directory names "@loader_path"
```
$ delocate-wheel dist/*

/usr/local/lib/python3.7/site-packages/delocate/libsana.py:110: UserWarning: Couldn't find @rpath/libClp.1.dylib on paths:
	/tmp/dist/@loader_path
  '\n\t'.join(realpath(path) for path in rpaths),
```

Same thing with:
```
          cmd LC_RPATH
      cmdsize 24
         path $ORIGIN (offset 12)
```

### LINKED ISSUE #90: delocate-wheel exits 'succesfully' after encountering potential errors.

HexDecimal:
Added more stuff on top of my previous PR's.  This adds a `DependencyTree` class which uses itself to resolve dependencies recursively and is a little easier to read its attributes than parse the `tree_libs` return value.  I tried to use it to update `tree_libs` itself but it turns out everything relies on its original behavior.

Tests fail unless I remove recursion and remove libraries with `@loader_path` in their install name.  That's the current state of things.  It might work better if I kept a backwards compatible `tree_libs` and started switching tools to use `DependencyTree` instead.

Closes #91, my previous PR.
Closes #37, closes #62, closes #90, open issues involving missing  `@loader_path` support.
#7 might be solved with additional commits to `delocate_listdeps.py`.

Still a lot to work on obviously.  I just made this PR to report the current status and so that it doesn't clobber #91 which I might need to return to if this PR doesn't turn out well.  I've only really gotten familiar with `libsana.py` at the moment.

I'll also need to add tests for `DependencyTree` itself.  I doubt its `__repr__` works.


HexDecimal:
After several iterations of refactoring I ended up with the functions: `get_dependencies`, `walk_library`, and `walk_directory` in the `libsana.py` module rather than a new class.  These can be used collect all dependencies of a library or directory of libraries while also handing possible issues like cyclic dependencies.

`delocate_path` was modified to use the new system.  It should support `@loader_path` and now rejects situations with missing dependencies.  `copy_recurse` is no longer used since dependencies are discovered all at once.

`tree_libs` seems to still be in use by `delocate_listdeps`, but I think this PR is already doing too much so I probably won't touch it right now.  Code which is still using `tree_libs` will show up in pytest as warnings.

Travis tests have been difficult to work with.  The way asserts are currently handled break pytests error reporting and should be updated to use assert statements directly.  The `set -vx` line in the Travis script adds a lot of uncollapsible output making the logs hard to navigate.

I'm still missing `@loader_path` tests.  Mostly it's hard for me to create the libraries for them.  A good test might be to have a library with an install path of `@rpath/libtarget.dylib` loaded from an adjacent library with an rpath of `@loader_path/`, this seems like a common setup.

HexDecimal:
`delocate_tree_libs` can now copy and modify libraries at the same time.  It can now work with a complete `lib_dict` where the dependencies referenced in it will be moved during the process.

Some functions from `pytest_tools` were replaced by asserts so that I could debug them.  I also moved `set -vx` in the Travis script so that I could read the tests easier.  I ran into a major issue with one of the tests where it seems like the test did not match the documentation (https://github.com/matthew-brett/delocate/pull/94/commits/0d54dd8dbae3424f1108c291e3d13bc3fefe37f9), I did my best to salvage that test and put it into a new test called `test_delocate_path_install_names`.

So far my plans to make a dependency tree traversal more strict haven't turned out well due to what I think are optional dependencies.  Right now I just log these cases as errors and ignore them.  Issues like this were ignored without output before now.

I now have a working version with recursive `@loader_path` support.  I've used my own library depending on SDL to test it with.

This is my current output.  The errors are from missing files on the runner.  It might be too verbose to list these system files which aren't going to be copied anyway, but it's weird to me how these files are detected as missing but then the library can still load in an isolated environment afterwards.  Only `hidapi` was needed for me, and that library was bundled within SDL.
```
delocate-wheel -v dist/*.whl
ERROR:delocate.libsana:/usr/lib/libiconv.2.dylib not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/GameController.framework/Versions/A/GameController not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/CoreHaptics.framework/Versions/A/CoreHaptics not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/Metal.framework/Versions/A/Metal not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/QuartzCore.framework/Versions/A/QuartzCore not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/AudioToolbox.framework/Versions/A/AudioToolbox not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/Frameworks/hidapi.framework/Versions/A/hidapi
ERROR:delocate.libsana:/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/Frameworks/hidapi.framework/Versions/A/hidapi
ERROR:delocate.libsana:/usr/lib/libSystem.B.dylib not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/Frameworks/hidapi.framework/Versions/A/hidapi
ERROR:delocate.libsana:/System/Library/Frameworks/Carbon.framework/Versions/A/Carbon not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/Cocoa.framework/Versions/A/Cocoa not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/CoreAudio.framework/Versions/A/CoreAudio not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/CoreVideo.framework/Versions/A/CoreVideo not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/ForceFeedback.framework/Versions/A/ForceFeedback not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/IOKit.framework/Versions/A/IOKit not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/usr/lib/libSystem.B.dylib not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/AppKit.framework/Versions/C/AppKit not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/ApplicationServices.framework/Versions/A/ApplicationServices not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/System/Library/Frameworks/Foundation.framework/Versions/C/Foundation not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/usr/lib/libobjc.A.dylib not found, requested by /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
ERROR:delocate.libsana:/usr/lib/libSystem.B.dylib not found, requested by /private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmpzjsze8hl/wheel/tcod/_libtcod.abi3.so
INFO:delocate.delocating:Copying library /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2 to .dylibs/SDL2
INFO:delocate.delocating:Copying library /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/Frameworks/hidapi.framework/Versions/A/hidapi to .dylibs/hidapi
INFO:delocate.delocating:Modifying install name in _libtcod.abi3.so from @rpath/SDL2.framework/Versions/A/SDL2 to @loader_path/.dylibs/SDL2
INFO:delocate.delocating:Modifying install name in .dylibs/SDL2 from @rpath/hidapi.framework/Versions/A/hidapi to @loader_path/hidapi
Fixing: dist/tcod-11.19.3.dev6-cp35-abi3-macosx_10_9_x86_64.whl
Copied to package .dylibs directory:
  /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/Frameworks/hidapi.framework/Versions/A/hidapi
  /Users/runner/work/python-tcod/python-tcod/dependencies/SDL2-2.0.14/SDL2.framework/Versions/A/SDL2
```

This might be finished, but I'm not certain.  I'll at least be using this branch for my own wheels for the moment.

matthew-brett:
So sorry to be slow to reply to this and your other PR.  I'm a university lecturer, and things are really busy at the moment.

This is a) very welcome, thank you so much and b) a lot to review.

And you've got your other very welcome and large PR over at #93 .

I wonder, would you have any interest in taking over the job of maintainer?  I'm happy to continue to help, of course, but the code-base is soon going to be more fresh and up to date in your memory than mine.

HexDecimal:
> So sorry to be slow to reply to this and your other PR. I'm a university lecturer, and things are really busy at the moment.

I made this PR since my own project suddenly had a requirement for `@loader_path`.  I've been installing this branch directly and haven't been affected by it not being merged.

> And you've got your other very welcome and large PR over at #93 .

This PR doesn't change as many files as #93, so merging that one first would make this one easier to review.

> I wonder, would you have any interest in taking over the job of maintainer? I'm happy to continue to help, of course, but the code-base is soon going to be more fresh and up to date in your memory than mine.

I don't have a Mac to test things on.  It's hard for me to check the tools or build test libraries and I rely on open-source CI for doing anything.  I think I could update the library tests using Actions but I don't think GitHub has an old enough MacOS runner to compile the older x86 binaries.

I planned on doing simple maintenance tasks through PRs and maybe adding simple features now that I'm knowledgeable with the library.  I lack the experience to solve MacOS specific issues and this PR took more effort because I couldn't perform tasks directly on a Mac.

I'd be willing to try and help as a co-maintainer but I'd be overwhelmed on my own.

matthew-brett:
Right - that's totally fair - I have access to quite a few Macs, and use them all the time, so it's much more reasonable for me to responsible for the library.   I'll do my best to get to these PRs quickly.  Thanks again.

matthew-brett:
Sorry - coming back to this now, in an attempt to clear the backlog.

Is there anything else you'd like to do on this PR?  Or should I review as-is?

@isuruf - did you try this branch as an alternative to gh-100?

HexDecimal:
It's been a very long time since I last checked this PR.  I couldn't merge this since I added too many type hints, but I needed those type hints in order to design this PR in the first place.  This PR was doomed since #93 couldn't be merged.

I can try rebasing this and removing any unnecessary changes if you want.

matthew-brett:
I've got time now to review this properly.   Would you prefer to put this one together with the type hints?  We can also drop Python 2, if that would help.

HexDecimal:
Dropping Python 2 would only help all future work.  It won't really affect what that's already been done.

I think I can go with whatever plan you want.  I feel like I can split my previous PR's up if necessary, or I can just resolve merge conflicts in this one.

matthew-brett:
Just checking - this PR includes the changes in #93?

HexDecimal:
> Just checking - this PR includes the changes in #93?

Yes.

matthew-brett:
OK - then please do just rebase, and I'll work through it, probably tomorrow.

isuruf:
> @isuruf - did you try this branch as an alternative to gh-100?

I haven't, but I'll try this branch soon.

HexDecimal:
I've already rebased this branch so I'm not sure if it still works until tests pass.  The TravisCI jobs are taking a while.

HexDecimal:
I've finished the rebase and the tests now pass again.  This is ready for review.

delocate/delocating.py
```diff
@@ -80,29 +91,51 @@ def delocate_tree_libs(lib_dict, lib_path, root_path):
...
+        logger.info(
+            "Copying library %s to %s", old_path, relpath(new_path, root_path)
+        )
+        shutil.copy(old_path, new_path)
+        # Delocate this file now that it is stored locally.
+        delocated_libs.add(new_path)
+        # Update lib_dict with the new file paths.
+        lib_dict[new_path] = lib_dict[old_path]
+        del lib_dict[old_path]
+        for required in list(lib_dict):
```
matthew-brett, line 112: Can't we just use

```suggestion
        for required in lib_dict:
```

?

delocate/delocating.py
```diff
@@ -80,29 +91,51 @@ def delocate_tree_libs(lib_dict, lib_path, root_path):
...
+    # Copy libraries outside of root_path to lib_path.
+    for old_path in copied_libs:
+        new_path = realpath(pjoin(lib_path, basename(old_path)))
+        logger.info(
+            "Copying library %s to %s", old_path, relpath(new_path, root_path)
+        )
+        shutil.copy(old_path, new_path)
+        # Delocate this file now that it is stored locally.
+        delocated_libs.add(new_path)
+        # Update lib_dict with the new file paths.
```
matthew-brett, line 109: Now modifying `lib_dict` in-place.  Can we fill another dictionary instead?  Or copy `lib_dict` to avoid modifying it in-place?   Not that anyone would likely notice, but just to keep the function behavior the same.

delocate/delocating.py
```diff
@@ -135,7 +168,16 @@ def copy_recurse(lib_path, copy_filt_func=None, copied_libs=None):
     copied_libs : dict
         Input `copied_libs` dict with any extra libraries and / or dependencies
         added.
+
+    .. deprecated:: 0.8
```
matthew-brett, line 172: Also - add note in `Changelog` ?

delocate/delocating.py
```diff
@@ -148,7 +190,12 @@ def copy_recurse(lib_path, copy_filt_func=None, copied_libs=None):
     return copied_libs
 
 
-def _copy_required(lib_path, copy_filt_func, copied_libs):
+def _copy_required(
```
matthew-brett, line 193: Note that this function will be orphaned by removal of deprecated `copy_recurse`, and removed with it?

delocate/delocating.py
```diff
@@ -235,18 +282,24 @@ def _copy_required(lib_path, copy_filt_func, copied_libs):
...
                 libname.startswith('/System'))
 
 
-def delocate_path(tree_path, lib_path,
-                  lib_filt_func=None,
-                  copy_filt_func=filter_system_libs):
+def delocate_path(
+    tree_path,  # type: Text
+    lib_path,  # type: Text
+    lib_filt_func=None,  # type: Union[None, str, Callable[[Text], bool]]
```
matthew-brett, line 299: Admittedly longer, but maybe more explicit as:

```suggestion
    lib_filt_func=None,  # type: Optional[Union[str, Callable[[Text], bool]]]
```

delocate/delocating.py
```diff
@@ -277,20 +330,33 @@ def delocate_path(tree_path, lib_path,
...
+    Raises
+    ------
+    DependencyNotFound
+        When any dependencies can not be located.
     """
     if lib_filt_func == "dylibs-only":
         lib_filt_func = _dylibs_only
+    elif isinstance(lib_filt_func, str):
+        raise TypeError('lib_filt_func string can only be "dylibs-only"')
+    lib_filt_func = lib_filt_func or (lambda _: True)
```
matthew-brett, line 343: Consider more explicit:

```suggestion
    lib_filt_func = lambda _ : True if lib_filt_func is None else lib_filt_func
```

delocate/delocating.py
```diff
@@ -277,20 +330,33 @@ def delocate_path(tree_path, lib_path,
...
-                        if copy_filt_func(key))
-    copied = delocate_tree_libs(lib_dict, lib_path, tree_path)
-    return copy_recurse(lib_path, copy_filt_func, copied)
+
+    lib_dict = {}  # type: Dict[Text, Dict[Text, Text]]
+    for library_path in walk_directory(tree_path, lib_filt_func):
+        for depending_path, install_name in get_dependencies(library_path):
+            if copy_filt_func and not copy_filt_func(depending_path):
+                continue
+            lib_dict.setdefault(depending_path, {})
```
matthew-brett, line 352: I guess you could do:

```
from collections import defaultdict

    lib_dict = defaultdict(dict)
```

to avoid this line in the loop.

matthew-brett (commented):
Some comments, working through.

.travis.yml
```diff
@@ -42,18 +42,18 @@ matrix:
...
+        (cd wheel_makers && sh -vx ./make_wheels.sh);  # echo commands
     fi
   - pip install .
 
 script:
   - echo $PWD
   - mkdir tmp_test_dir
   - cd tmp_test_dir
-  - pytest --pyargs delocate
+  - pytest --pyargs delocate --log-level DEBUG
```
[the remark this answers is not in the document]
HexDecimal: The logged output doesn't show up unless a test fails.  You can't test delocate locally without macOS so it'd have to be enabled again if I or anyone else needed to debug a test running on TravisCI.  That's the reason I left it in.

delocate/delocating.py
HexDecimal: I don't think I realized I modified it in-place.

I'll switch the `Dict` types to a `Mapping` type which will make the input parameter read-only, then copy the dictionary before modifying it.

delocate/delocating.py
HexDecimal: Yes, I messed up here and thought I was modifying the keys of `lib_dict` in-place, when that wasn't the case.

.travis.yml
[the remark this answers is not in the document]
matthew-brett: Fair enough.

delocate/delocating.py
HexDecimal: I assume I considered this.  I might have worried about how a defaultdict would behave when sent into other functions.

It's a lot to add to remove one line.  I prefer the current version.

matthew-brett: Fair enough.  I see what you mean.  The fix is just to convert it to a dict on the way out ...  Your call.

delocate/delocating.py
HexDecimal: At that point the type was `Optional[Callable[[Text], bool]]` so I decided that simple truth testing was okay.

You put the lambda at the start and removed the parentheses around it so the ternary is now part of the lambda.  You also included an odd whitespace after the underscore.

If I was focusing solely on readability I'd go with this:
```python
if lib_filt_func is None:
    lib_filt_func = lambda _: True
```
But this version does a better job of replacing the `or` operator:
```python
lib_filt_func = lib_filt_func if lib_filt_func is not None else lambda _: True
```

matthew-brett (commented):
Finish up the comments, first pass.

It would be good to have direct tests of the new functions:

* `walk_library`.
* `walk_directory`.

Are we testing `executable_path` somewhere?

Thanks - the type-hinting is useful, and the test changes are nice too.

matthew-brett:
Did you have any thoughts on the right way to close #7 ?

HexDecimal:
> Did you have any thoughts on the right way to close #7 ?

The new `libsana` tools in this PR should support this and might already be in effect, but it will be a flat list.

delocate/delocating.py
```diff
@@ -59,17 +71,45 @@ def delocate_tree_libs(lib_dict, lib_path, root_path):
...
+    _update_install_names(
+        lib_dict, root_path, libraries_to_delocate | copied_libraries
+    )
+    return libraries_to_copy
+
+
+def _analyze_tree_libs(
+    lib_dict,  # type: Mapping[Text, Mapping[Text, Text]]
+    root_path  # type: Text
+) -> Tuple[Dict[Text, Dict[Text, Text]], Set[Text]]:
```
matthew-brett, line 92: Nice split of the function - but I guess we are now depending on Python > 3.5.  That's fine, just noting.

delocate/tests/test_delocating.py
```diff
@@ -140,25 +150,43 @@ def test_delocate_tree_libs():
...
+
+def test_delocate_path_install_names():
+    # type: () -> None
+    # Test install name modifications.
+    with InTemporaryDirectory() as tmpdir:
+        subtree = pjoin(tmpdir, 'subtree')
+        all_local_libs = _make_libtree(subtree)
+        lib_path = "dynlibs"
+        copied = delocate_path(".", lib_path, copy_filt_func=None)
+        print(copied)
```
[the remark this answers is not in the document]
HexDecimal: These are debug prints in tests.  I left them in since they're hidden when tests pass.

I could remove them which would be reasonable since they're pretty vague.

delocate/delocating.py
HexDecimal: I did that by reflex.  I'll fix it since tests won't pass otherwise.

delocate/delocating.py
HexDecimal: Where do I add it?  May changelogs always have an 'unreleased' section for unreleased updates.

matthew-brett:
Yes - that's a good idea.  Would you mind adding an unreleased section at
the top in the same format as the other sections, with the notes?

delocate/libsana.py
```diff
@@ -4,16 +4,195 @@
...
+                logger.debug(
+                    "%s resolved to: %s", install_name, dependency_path
+                )
+        except DependencyNotFound:
+            logger.error(
+                "\n{0} not found:"
+                "\n  Needed by: {1}"
+                "\n  Search path:\n    {2}".format(
+                    install_name, lib_path, "\n    ".join(rpaths)
+                )
```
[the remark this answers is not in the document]
HexDecimal: f-strings are Python 3.6.  I'd be using them everywhere if I could.

delocate/libsana.py
```diff
@@ -48,73 +226,94 @@ def tree_libs(start_path, filt_func=None):
...
-
-    If the `lib_path` doesn't have `@rpath` then it's returned as is.
+def resolve_rpath(lib_path, rpaths, loader_path, executable_path="."):
+    # type: (Text, Iterable[Text], Text, Text) -> Text
+    """ Return `lib_path` with any special runtime linking names resolved.
 
     If `lib_path` has `@rpath` then returns the first `rpaths`/`lib_path`
-    combination found.  If the library can't be found in `rpaths` then a
-    detailed warning is printed and `lib_path` is returned as is.
+    combination found.  If the library can't be found in `rpaths` then
```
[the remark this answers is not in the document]
HexDecimal: Yes.  This lets you know about missing dependencies _before_ you deploy a wheel.

This is a critical issue, and not something that a should be ignored, and CI's don't care about warnings.

delocate/libsana.py
```diff
@@ -48,73 +226,94 @@ def tree_libs(start_path, filt_func=None):
...
+        The path to a library file, which may or may not be a relative path
+        starting with `@rpath`, `@loader_path`, or `@executable_path`.
     rpaths : sequence of str
         A sequence of search paths, usually gotten from a call to `get_rpaths`.
+    loader_path : str
+        The path to be used for `@loader_path`.
+        This must be the directory of the library which is loading `lib_path`.
+    executable_path : str, optional
+        The path to be used for `@executable_path`.
+        Assumed to be the working directory by default.
```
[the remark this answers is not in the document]
HexDecimal: It isn't a good default. It was hard to imagine what circumstances would lead to having to delocate a Python library using `@executable_path`, but I put something there anyway.

I can use `sys.base_exec_prefix` to make `@executable_path` relative to the Python executable. That seems like a reasonable choice.

matthew-brett:
I think this is a case for:

"Explicit is better than implicit."

and

"In the face of ambiguity, refuse the temptation to guess."

https://www.python.org/dev/peps/pep-0020/

My preference would be not to guess, but to error, unless the user passes
us the executable path they want.

delocate/libsana.py
```diff
@@ -4,16 +4,195 @@
...
+        The library to fetch dependencies from.  Must be an existing file.
+
+    Yields
+    ------
+    dependency_path : str
+        The direct dependencies of this library which can have dependencies
+        themselves.
+        If the libaray at `install_name` can not be found then this value will
+        not be a valid file path.
+        :func:`os.path.isfile` can be used to test if this file exists.
```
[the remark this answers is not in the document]
HexDecimal: That was there to remind me which function to use to check the results with.

delocate/libsana.py
```diff
@@ -4,16 +4,195 @@
...
+
+def get_dependencies(
+    lib_path,  # type: Text
+):
+    # type: (...) -> Iterator[Tuple[Text, Text]]
+    """Iterate over this libraries dependences.
+
+    Parameters
+    ----------
+    lib_path : str
```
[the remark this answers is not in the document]
HexDecimal: This means the path to the library.  Do you have an alternative name for it?

delocate/libsana.py
```diff
@@ -4,16 +4,195 @@
...
+        except DependencyNotFound:
+            logger.error(
+                "\n{0} not found:"
+                "\n  Needed by: {1}"
+                "\n  Search path:\n    {2}".format(
+                    install_name, lib_path, "\n    ".join(rpaths)
+                )
+            )
+            # At this point install_name is known to be a bad path.
+            # Expanding install_name with realpath may be undesirable.
```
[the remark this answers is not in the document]
HexDecimal: It was a (probably bad) attempt at keeping the return values consistent,  `realpath` gets used on every other return path in this function.

delocate/libsana.py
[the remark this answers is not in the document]
HexDecimal: On closer inspection the errors raised here are caught and reported by `get_dependencies`.  So this won't return a failure status on normal usage.

delocate/delocating.py
```diff
@@ -80,29 +121,81 @@ def delocate_tree_libs(lib_dict, lib_path, root_path):
...
+
+    Returns
+    -------
+    updated_lib_dict : dict
+        `lib_dict_` but modified so that dependices point to the libraries
+        at their destination.
+    needs_delocating : set of str
+        A set of the destination files, these need to be delocated.
+    """
+    lib_dict = {  # Make a clone of lib_dict.
```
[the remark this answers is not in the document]
matthew-brett: Easier to read as:

```
from copy import deepcopy

    lib_dict = deepcopy(lib_dict_)
```

?

delocate/delocating.py
```diff
@@ -277,38 +393,54 @@ def delocate_path(tree_path, lib_path,
...
 def _merge_lib_dict(d1, d2):
+    # type: (Dict[Text, Dict[Text, Text]], Mapping[Text, Mapping[Text, Text]]) -> None  # noqa: E501
     """ Merges lib_dict `d2` into lib_dict `d1`
     """
     for required, requirings in d2.items():
         if required in d1:
             d1[required].update(requirings)
         else:
-            d1[required] = requirings
+            d1[required] = dict(requirings)
```
matthew-brett, line 430: Why `dict` here?   The other branch does not copy...

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
 
 
-def tree_libs(start_path, filt_func=None):
+logger = logging.getLogger(__name__)
+
+
+class DependencyNotFound(Exception):
+    """Raised by tree_libs or resolve_rpath if an expected dependency is
+    missing.
+    """
```
matthew-brett, line 36: First line of docstring should be self-contained:

```suggestion
    """
    Raised by tree_libs or resolve_rpath if an expected dependency is missing.
    """
```

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+
+
+def get_dependencies(lib_path):
+    # type: (Text) -> Iterator[Tuple[Text, Text]]
+    """Iterate over this libraries dependences.
```
matthew-brett, line 41:
```suggestion
    """Iterator yielding dependencies of library `lib_fname`
```

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+logger = logging.getLogger(__name__)
+
+
+class DependencyNotFound(Exception):
+    """Raised by tree_libs or resolve_rpath if an expected dependency is
+    missing.
+    """
+
+
+def get_dependencies(lib_path):
```
matthew-brett, line 39:
```suggestion
def get_dependencies(lib_fname):
```

delocate/libsana.py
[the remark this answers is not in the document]
matthew-brett:
```suggestion
    lib_fname : str
```

delocate/libsana.py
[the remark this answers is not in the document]
matthew-brett: Still needed for user reading docstring?

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+        not be a valid file path.
+        :func:`os.path.isfile` can be used to test if this file exists.
+    install_name : str
+        The install name of `dependency_path` as if :func:`get_install_names`
+        was called.
+
+    Raises
+    ------
+    DependencyNotFound
+        When `lib_path` does not exist.
```
matthew-brett, line 63:
```suggestion
        When `lib_fname` does not exist.
```

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+    """
+    if not os.path.isfile(lib_path):
+        raise DependencyNotFound(lib_path)
+    rpaths = get_rpaths(lib_path) + get_environment_variable_paths()
+    for install_name in get_install_names(lib_path):
```
matthew-brett, line 68:
```suggestion
    if not os.path.isfile(lib_fname):
        raise DependencyNotFound(lib_fname)
    rpaths = get_rpaths(lib_fname) + get_environment_variable_paths()
    for install_name in get_install_names(lib_fname):
```

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+        try:
+            if install_name.startswith("@"):
+                dependency_path = resolve_rpath(
+                    install_name,
+                    rpaths,
+                    loader_path=dirname(lib_path),
```
matthew-brett, line 74:
```suggestion
                    loader_path=dirname(lib_fname),
```

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+            if dependency_path != install_name:
+                logger.debug(
+                    "%s resolved to: %s", install_name, dependency_path
+                )
+        except DependencyNotFound:
+            logger.error(
+                "\n{0} not found:"
+                "\n  Needed by: {1}"
+                "\n  Search path:\n    {2}".format(
+                    install_name, lib_path, "\n    ".join(rpaths)
```
matthew-brett, line 88:
```suggestion
                    install_name, lib_fname, "\n    ".join(rpaths)
```

delocate/libsana.py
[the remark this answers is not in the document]
matthew-brett: Yes, sorry, nothing to do here, just a thought for later.

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+    visited=None,  # type: Optional[Set[Text]]
+):
+    # type: (...) -> Iterator[Text]
+    """Iterate over all of this tree's dependencies inclusively.
+
+    Dependencies which can not be resolved will be logged and ignored.
+
+    Parameters
+    ----------
+    lib_path : str
```
matthew-brett, line 108: As above - I think this this would be easier to read as `lib_fname`.

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+        The library to start with.
+    filt_func : callable, optional
+        A callable which accepts filename as argument and returns True if we
+        should inspect the file or False otherwise.
+        Defaults to inspecting all files for library dependencies.
+        If `filt_func` filters a library it will also exclude all of that
+        libraries dependencies as well.
+    visited : set of str
```
matthew-brett, line 116:
```suggestion
    visited : None or set of str, optional
```

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+            yield realpath(install_name), install_name
+
+
+def walk_library(
+    lib_path,  # type: Text
+    filt_func=lambda filepath: True,  # type: Callable[[Text], bool]
+    visited=None,  # type: Optional[Set[Text]]
+):
+    # type: (...) -> Iterator[Text]
+    """Iterate over all of this tree's dependencies inclusively.
```
matthew-brett, line 102:
```suggestion
    """
    Yield all libraries on which `lib_path` depends, directly or indirectly.

    First yields `lib_path` itself, if not already `visited` and then all
    dependencies of `lib_path`, including dependencies of dependencies.
    """
```

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+        ):
+            yield sub_dependency
+
+
+def walk_directory(
+    root_path,  # type: Text
+    filt_func=lambda filepath: True,  # type: Callable[[Text], bool]
+):
+    # type: (...) -> Iterator[Text]
+    """Walk along dependences starting with the libraries within `root_path`.
```
matthew-brett, line 155:
```suggestion
    """Walk along dependencies starting with the libraries within `root_path`.
```

delocate/libsana.py
```diff
@@ -4,16 +4,194 @@
...
+    Parameters
+    ----------
+    root_path : str
+        The root directory to search for libraries depending on other libraries.
+    filt_func : None or callable, optional
+        A callable which accepts filename as argument and returns True if we
+        should inspect the file or False otherwise.
+        Defaults to inspecting all files for library dependencies.
+        If `filt_func` filters a library it will also exclude all of that
+        libraries dependencies as well.
```
matthew-brett, line 168:
```suggestion
        If `filt_func` filters a library it will will not further analyze any
        of that library's dependencies.
```

delocate/libsana.py
```diff
@@ -48,73 +225,104 @@ def tree_libs(start_path, filt_func=None):
...
+                lib_dict.setdefault(dependency_path, {})
+                lib_dict[dependency_path][depending_path] = install_name
     return lib_dict
 
 
-def resolve_rpath(lib_path, rpaths):
-    """ Return `lib_path` with its `@rpath` resolved
-
-    If the `lib_path` doesn't have `@rpath` then it's returned as is.
+def resolve_rpath(lib_path, rpaths, loader_path=None, executable_path=None):
```
matthew-brett, line 260: Doesn't this need a rename now?   Maybe something like `resolve_dep_paths`?

matthew-brett (commented):
Some more thoughts.  Thanks for the responses ...

delocate/delocating.py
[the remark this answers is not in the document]
HexDecimal: This is to convert read-only `Mapping[Text, Mapping[Text, Text]]` type into a modifiable `Dict[Text, Dict[Text, Text]]` type.  The copy functions will copy the read-only nature of the original type.  The more strict protocols are why `x.copy()` was replaced with `dict(x)` in some places.

`lib_dict` has a different type than `lib_dict_`.  Normally you could reuse `lib_dict` for the copy but I got `Dict[Text, Mapping[Text, Text]]` when I tried that.  This is why I renamed the parameter to `lib_dict_`.  This is likely an error with Mypy but I didn't look for a relevant bug report.

A new `_copy_lib_dict` function might be the best way to handle this.

delocate/delocating.py
```diff
@@ -59,17 +71,46 @@ def delocate_tree_libs(lib_dict, lib_path, root_path):
...
     rp_root_path = realpath(root_path)
-    rp_lib_path = realpath(lib_path)
-    # Test for errors first to avoid getting half-way through changing the tree
     for required, requirings in lib_dict.items():
-        if required.startswith('@'):  # assume @rpath etc are correct
-            # But warn, because likely they are not
-            warnings.warn('Not processing required path {0} because it '
-                          'begins with @'.format(required))
+        if required.startswith('@'):
+            # @rpath, etc, at this point is likely incorrect.
```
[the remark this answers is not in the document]
HexDecimal: `get_dependencies` can still return paths with `@` if they can't be resolved which is most likely to happen if a library dependency can't be found on any of the rpaths. If that happens then it's noted here and by `get_dependencies`.

So the question is should this be hard error rather than just logging it.  I think this should be treated like a missing library and raise `DelocationError`.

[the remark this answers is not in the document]
matthew-brett: Would you mind putting in a comment to say how the @ would reach this line?

For error vs logging - the question is - how to support the `delocate-listdeps` option, of showing all the library dependencies.  To be honest, I'm not sure what happens now if some of the depencies are missing, but I thought what happened is that the user gets warnings, but no error.  This is probably the right behavior for the `listdeps` option, but probably the wrong behavior for making a wheel - although it's possible someone somewhere is relying on missing dependencies passing silently.  We could make it a command line option, and a keyword argument to `tree_libs`, and `wheel_libs`.   Or make `_analyze_tree_libs` public and use that instead, with a keyword argument.  What do you think?

delocate/delocating.py
HexDecimal: `requirings` here is a `Mapping[Text, Text]` type and must be converted into a Dict to be compatible with assigning to `d1`.

The type hinting I added here revealed an edge-case where `d1` could be given references to values in `d2` which could mean changes made to `d1` could also affect `d2`.  I assumed that wasn't intended.

delocate/libsana.py
HexDecimal: This was an update to the already existing function, and if you took out rpaths then this would just be a simple string replace operation.

delocate/delocating.py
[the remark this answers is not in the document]
HexDecimal: > although it's possible someone somewhere is relying on missing dependencies passing silently. We could make it a command line option

I agree with making this a command line option just in case.

> For error vs logging - the question is - how to support the delocate-listdeps option, of showing all the library dependencies.

While it's related, I think `delocate-listdeps` is outside the scope of this PR, which was already so big that I had to cull a lot of changes to make it presentable.

I think the current tests assume the `delocate-listdeps` does not look up libraries recursively.  I had to leave `delocate-listdeps` using the old `tree_libs` code path for at the moment, but it should be possible to switch it to use `walk_directory` later.

delocate/delocating.py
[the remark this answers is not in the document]
matthew-brett: Thanks - that is clearer. 

How about:

```python
from typing import cast

    lib_dict = cast(Dict[Text, Mapping[Text, Text]], deepcopy(lib_dict_))
```

Yes it's a little bit verbose, but it does make the intention clear.

delocate/delocating.py
matthew-brett:
```suggestion
            # Convert to dict to satisfy type checking.
            d1[required] = dict(requirings)
```

delocate/libsana.py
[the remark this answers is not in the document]
matthew-brett:
```suggestion
```

delocate/libsana.py
matthew-brett:
```suggestion
    lib_fname : str
```

delocate/libsana.py
```diff
@@ -4,16 +4,198 @@
...
+        libraries dependencies as well.
+    visited : None or set of str, optional
+        We update `visited` with new library_path's as we visit them, to
+        prevent infinite recursion and duplicates.  Input value of None
+        corresponds to the set `{lib_path}`.  Modified in-place.
+
+    Yields
+    ------
+    library_path : str
+        The paths of each library including `lib_path` without duplicates.
```
matthew-brett, line 128:
```suggestion
    library_path : str
        The path of each library depending on `lib_fname`, including
        `lib_fname`, without duplicates.
```

delocate/libsana.py
```diff
@@ -4,16 +4,198 @@
...
+    """
+    if visited is None:
+        visited = {lib_path}
```
matthew-brett, line 131:
```suggestion
        visited = {lib_fname}
```

delocate/libsana.py
```diff
@@ -4,16 +4,198 @@
...
+    elif lib_path in visited:
```
matthew-brett, line 132:
```suggestion
    elif lib_fname in visited:
```

delocate/libsana.py
```diff
@@ -4,16 +4,198 @@
...
+        visited.add(lib_path)
+    if not filt_func(lib_path):
+        logger.debug("Ignoring %s and its dependencies.", lib_path)
+        return
+    yield lib_path
+    for dependency_path, _ in get_dependencies(lib_path):
+        if not os.path.isfile(dependency_path):
+            logger.error(
+                "%s not found, requested by %s", dependency_path, lib_path,
+            )
```
matthew-brett, line 144:
```suggestion
        visited.add(lib_fname)
    if not filt_func(lib_fname):
        logger.debug("Ignoring %s and its dependencies.", lib_fname)
        return
    yield lib_fname
    for dependency_fname, _ in get_dependencies(lib_fname):
        if not os.path.isfile(dependency_fname):
            logger.error(
                "%s not found, requested by %s", dependency_fname, lib_fname,
            )
```

delocate/libsana.py
```diff
@@ -4,16 +4,198 @@
...
+            continue
+        for sub_dependency in walk_library(
+            dependency_path,
```
matthew-brett, line 147: Just for consistency.

```suggestion
            dependency_fname,
```

delocate/libsana.py
matthew-brett: Right - yes - I'm just suggesting that string replace.  Maybe something like `resolve_special_paths` or similar, and then something like:

```python
def resolve_rpath(*args, **kwargs):
    """
    .. deprecated:: 0.8
        Please use `resolve_special_paths` instead of this function.
    """
    warnings.warn(
        "Please use `resolve_special_paths` instead of this function.",
        DeprecationWarning,
        stacklevel=2,
    )
    return resolve_special_paths(*args, **kwargs)
```

matthew-brett (commented):
Just a few more comments and suggestions.

Where are we with `executable_path`?

HexDecimal:
> Where are we with executable_path?

Removing `@executable_path` handling caused a regression with my own library.  If it raises an error to say that `executable_path` can't be handled then that aborts trying the remaining paths in rpaths.

matthew-brett:
> > Where are we with executable_path?
> 
> Removing `@executable_path` handling caused a regression with my own library. If it raises an error to say that `executable_path` can't be handled then that aborts trying the remaining paths in rpaths.

Yes, I agree it's a good idea to handle it.  If you want to leave it as-is for me to edit, that's fine.  But I still think it's a good idea to do it explicitly, as in:

```
delocate-wheel --executable_path=/usr/local/bin mywheel.whl
```

HexDecimal:
I'll add an option to set it manually, but I'll set the default path to the Python executable which will mimic how the libraries would be loaded normally in Python.

delocate/libsana.py
HexDecimal: I'm not sure what to do about it.  Trying to resolve rpaths without including `@loader_path` is fundamentally broken, the old function did not support loader_path and I shouldn't be making loader_path optional in the updated function.

I'll try making it a 2nd function.

delocate/libsana.py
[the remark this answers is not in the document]
HexDecimal: The unusual documentation was a hint at the real problem, which is that this function should have never returned a string for invalid paths at all.

delocate/libsana.py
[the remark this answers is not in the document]
HexDecimal: `dirname(sys.executable)` is now the default for `@executable_path`, and this can be changed from the command line.

delocate/delocating.py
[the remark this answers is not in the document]
HexDecimal: Requirements starting with `@` can no longer reach that section of the code.  They are always filtered out or resolved by other functions before going into `lib_dict`.

`get_dependencies` does not return invalid paths anymore, and it's up to callers of `get_dependencies` to handle what to do when libraries are missing.

Dependencies can now be ignored from the command line.  This is handled by the `delocate_path` function.

delocate/delocating.py
```diff
@@ -80,29 +121,81 @@ def delocate_tree_libs(lib_dict, lib_path, root_path):
...
-            set_install_name(requiring, orig_install_name,
-                             '@loader_path/{0}/{1}'.format(
-                                 req_rel, basename(required)))
-    for required in delocated_libs:
+            needs_delocating.add(required)
+    return needs_copying, needs_delocating
+
+
+def _copy_required_libs(
+    lib_dict_,  # type: Mapping[Text, Mapping[Text, Text]]
```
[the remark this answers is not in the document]
HexDecimal: Adding a function to copy lib_dict fixed the type issues.  No more underscore needed.

delocate/delocating.py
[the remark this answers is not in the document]
HexDecimal: All `Dict`'s are `Mapping`'s, but not all `Mapping`'s are `Dict`'s.  I fixed the type issues by adding a new function to copy lib_dict's.

delocate/libsana.py
```diff
@@ -48,73 +251,108 @@ def tree_libs(start_path, filt_func=None):
...
     lib_path : str
-        The path to a library file, which may or may not start with `@rpath`.
+        The path to a library file, which may or may not be a relative path
+        starting with `@rpath`, `@loader_path`, or `@executable_path`.
     rpaths : sequence of str
         A sequence of search paths, usually gotten from a call to `get_rpaths`.
+    loader_path : str, optional
+        The path to be used for `@loader_path`.
+        This should be the directory of the library which is loading `lib_path`.
+    executable_path : str, optional
```
matthew-brett, line 313:
```suggestion
    executable_path : None or str, optional
```

delocate/libsana.py
```diff
@@ -4,16 +4,220 @@
...
+    root_path : str
+        The root directory to search for libraries depending on other libraries.
+    filt_func : None or callable, optional
+        A callable which accepts filename as argument and returns True if we
+        should inspect the file or False otherwise.
+        Defaults to inspecting all files for library dependencies.
+        If `filt_func` filters a library it will will not further analyze any
+        of that library's dependencies.
+    executable_path : str, optional
+        An alternative path to use for resolving `@executable_path`.
```
matthew-brett, line 191:
```suggestion
        If not None, an alternative path to use for resolving `@executable_path`.
```

delocate/delocating.py
```diff
@@ -197,6 +305,9 @@ def _copy_required(lib_path, copy_filt_func, copied_libs):
     ``/sys/libB.dylib``, so we fix our copy of `libC.dylib`` to point to
     ``my_lib_path/libB.dylib`` and add ``/sys/libC.dylib`` as a
     ``dependings_dict`` entry for ``copied_libs['/sys/libB.dylib']``
+
+    .. deprecated:: 0.8
```
matthew-brett, line 309: Actually - 0.9 - no?   0.9 seems like a good next release number.

delocate/delocating.py
```diff
@@ -266,6 +385,10 @@ def delocate_path(tree_path, lib_path,
         Default is callable rejecting only libraries beginning with
         ``/usr/lib`` or ``/System``.  None means copy all libraries. This will
         usually end up copying large parts of the system run-time.
+    executable_path : str, optional
+        An alternative path to use for resolving `@executable_path`.
+    ignore_missing : bool, default=False
+        Continue even if missing dependencies are detected.
```
matthew-brett, line 391:
```suggestion

    executable_path : None or str, optional
        If not None, an alternative path to use for resolving `@executable_path`.
    ignore_missing : {False, True}, optional
        Continue even if missing dependencies are detected.
```

delocate/tests/test_libsana.py
```diff
@@ -172,15 +185,31 @@ def test_wheel_libs():
...
+
+def test_get_dependencies():
+    # type: () -> None
+    with pytest.raises(DependencyNotFound):
+        list(get_dependencies("nonexistent.lib"))
+    # This check might be too platform specific.
+    assert dict(get_dependencies(LIBA)) == {
+        realpath("/usr/lib/libstdc++.6.dylib"): "/usr/lib/libstdc++.6.dylib",
+        realpath("/usr/lib/libSystem.B.dylib"): "/usr/lib/libSystem.B.dylib",
+    }
```
matthew-brett, line 215: Consider direct tests for new functions:

* `walk_library`
* `walk_directory`

matthew-brett (commented):
More comments, nearly there.

I like where you are going with simplifying the checking for missing dependencies to None - I'm sure you're right to do that.

delocate/delocating.py
HexDecimal: `bool, default=False` is less ambiguous than `{False, True}, optional` and seemed valid according to this style guide:  https://numpydoc.readthedocs.io/en/latest/format.html#parameters

matthew-brett: Aha - the recommendation changed since the old days.   That's fine.

HexDecimal:
It's hard for me to set up tests.  I still can't run tests locally.  When I run tests remotely I have to wait for TravisCI which is sometimes slow.

There are no tests for `@loader_path`, and there are no tests for recursive dependencies.  I can't make the libraries for these right now.  `@executable_path` is a very low priority for me.

I can write other tests but it takes a long time for me to write them.

matthew-brett:
How about dumping a copy of your own undelocated wheel in the test data directory?  Too large?   Code that can't be shared?

delocate/delocating.py
```diff
@@ -80,29 +124,79 @@ def delocate_tree_libs(lib_dict, lib_path, root_path):
...
+
+    Returns
+    -------
+    updated_lib_dict : dict
+        A copy of `lib_dict` modified so that dependencies now point to
+        the copied library destinations.
+    needs_delocating : set of str
+        A set of the destination files, these need to be delocated.
+    """
+    lib_dict = _copy_lib_dict(lib_dict)
```
matthew-brett, line 152: I prefer the casting solution by a fair margin - as in:

```
lib_dict_out = cast(
        Dict[Text, Mapping[Text, Text]],
        deepcopy(lib_dict))
```

Why?   Because it makes it clear _why_ we need to do this - which is a) to avoid stamping on the input dictionary (the `deepcopy`) and b) to satisfy the type-system (`cast`).   At the moment, I think someone reading the code will think what I thought, which is "Hmm, that's funny, I wonder why they are doing that".  Of course we could answer it with comments, but the `cast` code above answers it immediately, without need for a new function to look at.

HexDecimal:
My wheel depends on SDL and the build process for it isn't great.  It has to be delocated with files downloaded as part of the build step.  It's around 2 MB total.

The thing which required me to make this PR was that SDL added a dependency which needed to be included.  It used `@rpath` with `@loader_path` to load the other library from next to itself.  This should be reproducible with install_name and rpath flags when building the test libraries.

I can make macOS build artifacts on GitHub Actions, but that doesn't support the `i386` architecture.  So I can't edit the existing build scripts.

matthew-brett:
It's fine to drop i386 and replace with arm64.  Do you want me to do that?  I can put in a separate PR and you can rebase?

delocate/delocating.py
HexDecimal: I keep pushing back on this idea because `cast` isn't doing what you think it's doing.  In this case a bad cast is harmless because it's a private function and we know for sure the actual type being passed in, but if a Mapping type which isn't a Dict is ever passed in then this cast will break code depending on the types being dict.  The methods expected in Python's dict type such as `dict.update` would be missing for example.

Also note that `lib_dict_out` has to be `Dict[Text, Dict[Text, Text]]`.  I mentioned `Dict[Text, Mapping[Text, Text]]` as an undesirable type I was getting before when Mypy didn't recognize what I was trying to do.

If I change the input type of `lib_dict` to be `Dict[Text, Dict[Text, Text]]` then `deepcopy` will work how you expect it to without needing any kind of cast or trick.  This would be okay to do on private function.  Any comments or name changes can also be added to make it more clear, but I wanted to get rid of the original variable so that it couldn't be tampered with accidentally.

HexDecimal:
Dropping i386 means I could try messing with the make libs scripts.  I think I can setup the GitHub workflows for it if you did that.

delocate/delocating.py
matthew-brett: I did know what cast was doing, but it's true I hadn't considered non-dict types as input values.   I still think it would be more readable if the copy has a different name from the original - and some comments here would be good, so that no-one else finds themselves in the same dark corner as me ...  On the lines of:

```
# Make copy of `lib_dict` where mapping and mapping values are dict copies.
# This prevents us overwriting the input `lib_dict` and enforces mutable dicts
# in the working variable and output.
```

delocate/libsana.py
```diff
@@ -48,54 +252,131 @@ def tree_libs(start_path, filt_func=None):
...
+                        install_name
+                    )
+                    continue
+                lib_dict.setdefault(dependency_path, {})
+                lib_dict[dependency_path][depending_path] = install_name
     return lib_dict
 
 
+def resolve_dynamic_paths(lib_path, rpaths, loader_path, executable_path=None):
+    # type: (Text, Iterable[Text], Optional[Text], Optional[Text]) -> Text
```
matthew-brett, line 293: `loader_path` no longer optional?

delocate/libsana.py
```diff
@@ -48,54 +252,131 @@ def tree_libs(start_path, filt_func=None):
...
+    Raises
+    ------
+    DependencyNotFound
+        When `lib_path` has `@rpath` in it but no library can be found on any
+        of the provided `rpaths`.
+    """
+    if executable_path is None:
+        executable_path = dirname(sys.executable)
+    if lib_path.startswith('@loader_path/'):
+        if loader_path is None:
```
matthew-brett, line 331: `loader_path` still allowed to be None?

matthew-brett (commented):
A couple of comments.

delocate/libsana.py
HexDecimal: My mistake.  I've removed the Optional type and this check.

delocate/libsana.py
HexDecimal: It's important that `loader_path` has a correct value.  Making it optional causes issues that would be difficult to debug.  I made it optional before to be compatible with the previous function, but now that it's its own function it will no longer be optional.

An optional `loader_path` or `executable_path` would cause paths in a libraries rpaths to become invalid.  This causes issues with the order of the search paths.  I've been trying to avoid silently skipping search paths.

matthew-brett: Sorry - I wasn't clear - was only trying to point out that the type checking didn't match any more - totally agree about making it not-optional.

matthew-brett:
I've put a PR in to update to Big Sur, replace i386 with arm64 for dual architecture : #111

matthew-brett:
Would you consider doing a rebase after the recent changes?   Happy to do it if you have not got time.

delocate/tests/test_libsana.py
```diff
@@ -173,15 +187,43 @@ def test_wheel_libs():
...
     # Should return the given parameter as is since it can't be found
     assert_equal(resolve_rpath(lib_rpath, []), lib_rpath)
+
+
+def test_get_dependencies():
+    # type: () -> None
+    with pytest.raises(DependencyNotFound):
+        list(get_dependencies("nonexistent.lib"))
+    # This check might be too platform specific.
+    assert dict(get_dependencies(LIBA)) == {
```
matthew-brett, line 226: Could use `test_install_names.EXT_LIBS` for this, to allow some indirection.

HexDecimal:
I'm having test failures in the TravisCI build wheels job and I'm not sure why.  It's unusual that all other jobs seem to pass except for that one.

matthew-brett:
It looks like the test run is picking up the system `/usr/lib` libraries as dependencies - and they don't exist on the Big Sur images, because they are in something called the dynamic linker cache - they don't exist on the file-system.

Is the filtering not working correctly?

matthew-brett:
A pull request into your branch to fix an error I was getting testing locally: https://github.com/HexDecimal/delocate/pull/2

Changelog
```diff
@@ -14,6 +14,13 @@ lot of work discussing OSX things with MB, and contributes fixes too.
 Releases
 ********
 
+* Unreleased
+
+  * Libraries depending on ``@loader_path`` are now supported.
```
matthew-brett, line 19: Add note on `executable_path`?

matthew-brett (commented):
Have you tested on your own wheel as well?

Otherwise - I think this is good to go.  Do you agree?

HexDecimal:
> Have you tested on your own wheel as well?

I modified the rpath test wheel to add an indirect dependency very similar to what SDL has.  I think my wheels should be okay as long as that test passes.

matthew-brett:
@isuruf - we're planning to merge this branch soon - can you test it for your needs and let us know?

isuruf:
I checked this PR and with https://github.com/isuruf/delocate/commit/6717410bca3afbde3f98a9176e1b0f2369bc72b7 added on top of this PR, numpy arm64 wheels work correctly.

matthew-brett:
@HexDecimal - a couple of test failures on Travis - would you mind taking a look?

matthew-brett:
OK - let's give this a try.  Thanks @HexDecimal for all your patient work.

PR #96, opened 2021-02-24
Respect namespace_packages when searching for packages

rowillia:
Namespace packes do not have a top level `__init__.py` - https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages
This causes delcoate to fail to find those packages.

setuptools will store the those values in `*.dist-info/namespace_packages.txt`
https://github.com/pypa/setuptools/blob/main/setup.py#L139

https://github.com/matthew-brett/delocate/pull/38 blindly globbed everything, this approaches it in a more
compliant way.

### DIFF
--- a/delocate/tests/test_tools.py
+++ b/delocate/tests/test_tools.py
@@ -176,7 +176,8 @@ def test_find_package_dirs():
         a_dir = pjoin('to_test', 'a_dir')
         b_dir = pjoin('to_test', 'b_dir')
         c_dir = pjoin('to_test', 'c_dir')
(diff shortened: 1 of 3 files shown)


rowillia:
@matthew-brett Mind taking a look at this?

matthew-brett:
I'm very sorry - I'm mid marking crunch - I should have a lot more time starting next week.

rowillia:
👋 @matthew-brett any luck?

matthew-brett:
Sorry to be quiet - would you mind testing @HexDecimal's new PR at #123 ?  Does it work for your namespace packages?

PR #97, opened 2021-02-24
Enforce Flake8 with GitHub Actions.

HexDecimal:
Enables workflows which verify that Flake8 passes, and fixes that issues that already exist.

The config file at `.flake8` is pretty basic.  This check won't do as much until that's updated.

### DIFF
--- a/.github/workflows/python-linters.yml
+++ b/.github/workflows/python-linters.yml
@@ -1,0 +1,23 @@
+name: Python linters
+
+on:
+  push:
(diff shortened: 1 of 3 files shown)


rowillia:
@HexDecimal if you rebase on master tests should pass.  Otherwise looks good.

HexDecimal:
Rebased and tests have passed.

matthew-brett:
Thanks - in it goes.

PR #99, opened 2021-02-25
Enforce type annotations.

HexDecimal:
This adds checks for MyPy, and ensures the `typing` and `typing_extensions` modules are available. `py.typed` marks the package as typed.

`six` wasn't added but it probably should be.  Python 2 string typing will be difficult without it.


matthew-brett:
@HexDecimal - anything left here that didn't go in with #94?

HexDecimal:
This PR still does a lot of important things, including adding Mypy to a GitHub workflow and marking Delocate as a type-checked package for scripts which import it.

I've just updated it for Python 3.6.

matthew-brett:
Feel free to merge when tests pass.

HexDecimal:
I'm not a maintainer with merge permissions.

matthew-brett:
Sorry - I forgot.  I've sent you an invite.

Issue #71, opened 2020-04-28
[BUG] When delocate is called with a non-expanded wildcard path it produces wrong result


webknjaz:
Imagine you call `python -m delocate.cmd.delocate_listdeps --all dist/*.whl` in the env where `dist/*.whl` is not expanded by the shell so it's effectively `python -m delocate.cmd.delocate_listdeps --all 'dist/*.whl'`. And the dists dir has:
```console
-rw-r--r--   1 runner  staff   144K Apr 28 08:30 ansible_pylibssh-0.0.1.dev1-cp38-cp38-macosx_10_9_x86_64.whl
```
It'll use `dist/ansible_pylibssh-0.0.1.dev1-cp38-cp38-macosx_10_9_x86_64.whl` as a source and will emit a wheel with the name `dist/*.whl`:
```console
-rw-r--r--   1 runner  staff   1.4M Apr 28 08:30 *.whl
-rw-r--r--   1 runner  staff   144K Apr 28 08:30 ansible_pylibssh-0.0.1.dev1-cp38-cp38-macosx_10_9_x86_64.whl
```
which then makes `twine check dist/*.whl` explode: https://github.com/pypa/twine/issues/612.

It looks like delocate applies glob in one place but not another which is a bug.

webknjaz:
UPD: It's worse now — recently, `delocate` started erroring out early with that glob being passed. It worked a few weeks ago so the changes are related to v0.10.5 or v0.10.6. Though, skimming through the diffs, it's unobvious what exactly caused this.

webknjaz:
I think, it's v0.10.5, after all: the last successful nightly build we had was on Nov 13 and since Nov 14, the macOS builds started failing. This correlates with the release date of v0.10.5 which happened on Nov 14, right in-between these CI cron runs.

webknjaz:
I addressed this by expanding the wildcard outside the delocate invocation: https://github.com/ansible/pylibssh/commit/025166d726ca669467451cbe3d9b603f7dc00c66.

HexDecimal:
@webknjaz What's the regression caused by v0.10.5? I mean what's the error output?

If the shell doesn't the expand the paths then delocate and probably most CLI's will take the path as-is. I don't see this as unexpected behavior but I don't think anyone would mind if delocate globbed the paths itself.

Should also add tests related to globbing.

HexDecimal:
[...]

PR #106, opened 2021-05-26
Add --exclude option to delocate-wheel

rossant:
Same motivation as https://github.com/pypa/auditwheel/pull/310

I need this to remove libvulkan from the repaired wheel.

### DIFF
--- a/delocate/cmd/delocate_wheel.py
+++ b/delocate/cmd/delocate_wheel.py
@@ -15,6 +15,9 @@ from os.path import join as pjoin
 from typing import List, Optional, Text
 
 from delocate import __version__, delocate_wheel
+from delocate.delocating import filter_system_libs
+
+logger = logging.getLogger(__name__)
 
 
 def main() -> None:
@@ -63,6 +66,16 @@ def main() -> None:
                 "--dylibs-only",
                 action="store_true",
                 help="Only analyze files with known dynamic library extensions",
+            ),
+            Option(
+                "-e",
+                "--exclude",
+                action="append",
+                default=[],
+                type="string",
+                help=(
+                    "Exclude any libraries where path includes the given string"
+                ),
             ),
             Option(
                 "--require-archs",
@@ -117,6 +130,21 @@ def main() -> None:
     else:
         require_archs = opts.require_archs
     lib_filt_func = "dylibs-only" if opts.dylibs_only else None
+
+    exclude_files: List[str] = opts.exclude
+
+    def copy_filter_exclude(name: str) -> bool:
+        """Returns False if name is excluded, uses normal rules otherwise."""
+        for exclude_str in exclude_files:
+            if exclude_str in name:
+                logger.info(
+                    "%s excluded because of exclude %r rule.",
+                    name,
+                    exclude_str,
+                )
+                return False
+        return filter_system_libs(name)
+
     for wheel in wheels:
         if multi or opts.verbose:
             print("Fixing: " + wheel)
@@ -129,6 +157,7 @@ def main() -> None:
             out_wheel,
             lib_filt_func=lib_filt_func,
             lib_sdir=opts.lib_sdir,
+            copy_filt_func=copy_filter_exclude,
             require_archs=require_archs,
             executable_path=opts.executable_path,
             ignore_missing=opts.ignore_missing_dependencies,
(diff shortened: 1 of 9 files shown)


delocate/cmd/delocate_wheel.py
```diff
@@ -59,6 +69,15 @@ def main():
...
         require_archs = opts.require_archs
     lib_filt_func = 'dylibs-only' if opts.dylibs_only else None
+
+    exclude = [
+        _.strip() for _ in (opts.exclude or '').split(',') if _.strip()]
+
+    def copy_filt(name):
+        if not exclude:
+            return filter_system_libs(name)
+        return filter_system_libs(name) and not _is_in_list(name, exclude)
```
matthew-brett, line 79: I think you are doing a substring search here?  Is that what you intended?  Or did you mean

```
and not name in exclude
```

?

delocate/cmd/delocate_wheel.py
```diff
@@ -37,6 +45,8 @@ def main():
                action="store_true",
                help="Only analyze files with known dynamic library "
                "extensions"),
+        Option("-e", "--exclude",
```
matthew-brett, line 48: Or allow multiple `--exclude` options with `action="append"`?

matthew-brett (commented):
Sorry for the delay.  Some thoughts.   Can you think of a test?

delocate/cmd/delocate_wheel.py
HexDecimal: The substring search by rossant is intentional.  The main alternative is to allow RegEx, which is usually overkill.

delocate/cmd/delocate_wheel.py
HexDecimal: I've rewritten the code to use `action="append"` but I haven't tested it.

matthew-brett (commented):
All looks good - but it still needs some tests - maybe in `delocate/tests/test_scripts.py` ?

HexDecimal:
Maybe.  I wish the script running tests were using a standard tool rather than `ScriptRunner`.  I'll probably come back to this later.

matthew-brett:
Yes, I wrote my own script testing framework around 2016, before there were others available - or at least - before I knew there were others available.

HexDecimal:
Won't merge until tests are made which is a reasonable request.  The new tests can use `ScriptRunner` like the current tests, or `pytest-console-scripts` which is more standard.

PhilipGarnero:
Can we merge this please ?
I know it's untested but it's still better than nothing. It's just one click away from being useful to people 🙇‍♂️

PhilipGarnero:
I took a few minutes to write something that should do ok as a test :

```py
@pytest.mark.xfail(sys.platform != "darwin", reason="Needs macOS linkage.")
def test_fix_wheel_with_excluded_dylibs():
    with InTemporaryDirectory() as tmpdir:
        fixed_wheel, stray_lib = _fixed_wheel(tmpdir)
        _rename_module(fixed_wheel, "module.other", "test.whl")
        shutil.copyfile("test.whl", "test2.whl")
        # We exclude the stray library so it shouldn't be present in the wheel
        code, stdout, stderr = run_command(["delocate-wheel", "-e", "extfunc", "test.whl"])
        with InWheel("test.whl"): 
            dylibs = pjoin("plat_pkg", "fakepkg1", ".dylibs")
            assert_equal(os.listdir(dylibs), [])
        # We exclude a library that does not exist so we should have a normal behavior
        code, stdout, stderr = run_command(["delocate-wheel", "-e", "doesnotexist", "test2.whl"])
        _check_wheel("test2.whl", ".dylibs")
```

HexDecimal:
I've added the test and cleaned up the commits. It could be better but now I think it's good enough to merge.

HexDecimal (approved):

HexDecimal:
Moved some CLI logic to a shared module so that any new code can apply to both `delocate-wheel` and `delocate-path` without having to copy that code across both modules.

webknjaz:
> Moved some CLI logic to a shared module so that any new code can apply to both `delocate-wheel` and `delocate-path` without having to copy that code across both modules.

Hey @HexDecimal, I think the changes in this PR may have resulted in breaking the delocate CLI extending a glob-style paths arg: https://github.com/matthew-brett/delocate/issues/71#issuecomment-1830984185.

PR #107, opened 2021-06-11
Updated URL

radarhere:
Navigating to http://pypi.python.org/pypi/delocate in your browser, you are redirected to https://pypi.org/project/delocate/


matthew-brett:
Thanks !

PR #112, opened 2021-07-09
Add GitHub workflow to build test data and run tests.

HexDecimal:
With the `i386` files dropped I was finally able to finish this.

This workflow rebuilds the test data and uploads it as an artifact, which means I can commit it back into the repository on my own.  It also runs the standard tests to verify that they work.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,50 @@
+# Builds wheels and libraries used to run tests, uploads them, then runs the standard tests.
+name: Build and run tests
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  tests:
+    runs-on: macos-10.15
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Build libs and wheels
+        # Use a venv from multibuild for this step.
+        run: |
+          source multibuild/osx_utils.sh
+          get_macpython_environment $MB_PYTHON_VERSION venv
+          make
+        env:
+          MB_PYTHON_VERSION: "3.9"
+          MB_PYTHON_OSX_VER: "10.9"
+      - name: Upload test data
```
matthew-brett, line 34: Should we upload after the tests?

HexDecimal: The upload is so the test data can be viewed.  I might want to look at that it even if the tests fail.

matthew-brett: Fair enough ...

Issue #109, opened 2021-07-03
Proposal - drop Pythons < 3.6


matthew-brett:
I propose that we drop all end-of-life Python releases for the master branch.  The oldest currently supported version is Python 3.6 - https://devguide.python.org/#status-of-python-branches

I don't think we have any reason to support very old Pythons - most users of delocate are release managers who will very likely have access to multiple versions of Python.

Any objections?

HexDecimal:
I'd also suggest dropping i386 architectures from the tests.  Support for that architecture has been completely dropped by MacOS as they've moved on to ARM64.  As hard as it is to setup a MacOS environment it's much harder that it also has to be a super old one as well.

matthew-brett:
Yes, dropping i386 is very sensible.

matthew-brett:
i386 dropped as of merge of #111 

Signal that we are dropping Python < 3.6 in #113.

Issue #104, opened 2021-05-02
New release


keichi:
Could you please publish a new release? I currently have to install delocate from GitHub because my extension relies on `libstdc++` and https://github.com/matthew-brett/delocate/pull/76 is needed.

matthew-brett:
Yes - sorry - overwhelmed at the moment - can you remind me early next
week, just in case I forget?

keichi:
Of course, thanks!

matthew-brett:
@HexDecimal - do you have a preferred email address to go in the AUTHORS file for the release?

HexDecimal:
> @HexDecimal - do you have a preferred email address to go in the AUTHORS file for the release?

`4b796c65+github@gmail.com` is preferred.  My name should be updated from `Stewart` to `Kyle Benesch`.

matthew-brett:
See #116 .

matthew-brett:
Done - please test and confirm it's working?

keichi:
Yes, 0.9.0 is working. Thanks!

PR #113, opened 2021-07-10
Drop claim to support Python < 3.6

matthew-brett:
Drop automated testing for Python < 3.6.

When this PR is merged, new code can use Python 3.6 features such as type annotation, 
f-strings and insertion-order dictionaries.

I suggest we strip out the Python 2.7 back-compatibility code gradually, as we do other 
edits on the code.

Closes gh-109.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ env:
 
 matrix:
   include:
-      - env: MB_PYTHON_VERSION=2.7
-      - env: MB_PYTHON_VERSION=3.5
       - env: MB_PYTHON_VERSION=3.6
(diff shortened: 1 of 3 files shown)


HexDecimal:
Don't forget to add `python_requires='>=3.6',` to delocate's `setup` function. https://packaging.python.org/guides/dropping-older-python-versions/

matthew-brett:
Any more comments here?   I will merge in a couple of days unless I hear otherwise.

HexDecimal:
You could perhaps make a final Python 2 release to deal with #104 before merging this.

PR #116, opened 2021-07-17
MRG: Updates ready for release

matthew-brett:
Anyone / anything I missed?


HexDecimal:
Note that you should keep the "Unreleased" section when updating the changelog.  You add the released version header between the Unreleased header and the contents, so that there's a new section to add the next changes to.

In general I try to conform to this standard: https://keepachangelog.com/en/1.0.0/

matthew-brett:
Will add back the uneleased section in the post-release commit.  Merging for now, but let me know if there's anything else y'all spotted.

PR #117, opened 2021-07-21
Update Flake8 config and workflow.

HexDecimal:
This sets which error codes to select and ignore.  The previous config was using parts of my local configuration which was undesirable.

The GitHub workflow now uses a simpler problem matcher action and was reformatted.

Any new Flake8 errors were fixed.

### DIFF
--- a/delocate/delocating.py
+++ b/delocate/delocating.py
@@ -346,16 +346,12 @@ def _copy_required(
         copied_libs[required] = procd_requirings
 
 
-def _dylibs_only(filename):
-    # type: (Text) -> bool
-    return (filename.endswith('.so') or
-            filename.endswith('.dylib'))
+def _dylibs_only(filename: str) -> bool:
+    return filename.endswith('.so') or filename.endswith('.dylib')
 
 
-def filter_system_libs(libname):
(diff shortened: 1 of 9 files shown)


delocate/delocating.py
```diff
@@ -348,14 +348,12 @@ def _copy_required(
 
 def _dylibs_only(filename):
     # type: (Text) -> bool
-    return (filename.endswith('.so') or
-            filename.endswith('.dylib'))
+    return (filename.endswith('.so') or filename.endswith('.dylib'))
```
matthew-brett, line 351: I was surprised by this needed change; which rule is this?

I think we don't need the outer parentheses any more.

matthew-brett (commented):
Just one comment / question.  Looks good.  Feel free to merge with or without response to comment.

delocate/delocating.py
HexDecimal: The `W503,W504` codes disallow line-breaks before/after operators.  I fixed this section before deciding to ignore the rule.

matthew-brett:
Please do merge when the tests are done ...

PR #118, opened 2021-07-22
Show recursive dependencies in listdeps command.

### LINKED ISSUE #7: Recursive dependencies not shown in delocate-listdeps

HexDecimal:
I finally got around to updating listdeps.

Added the `tree_libs_from_directory` function instead of duplicating this code.  This is an improved version of the `tree_libs` function.

The `wheel_libs` function was modified to use `tree_libs_from_directory`.
This is not fully backwards compatible since the new behavior will now raise an error instead of returning an incomplete lib dict.

The listdeps test now moves the working directory to the test data dir to get more consistent results.

Closes #7


HexDecimal:
I've changed the way the platform wheel is tested.  No more `wheel_build_path.txt`, now a custom fixture sets up a platform wheel depending on the external library from the test data directory.

The updated API understands that the external library can depend on libraries itself.  So tests needed to be edited to include that extra information.

matthew-brett:
Just checking quickly - this doesn't mean that we only test the code built on the platform on which we are testing?

HexDecimal:
I haven't changed which platforms run tests.  This still includes the tests that run without rebuilding the test data. I could now try running these tests in full isolation if you like.

`fakepkg1` now starts with the install name `libextfunc.dylib` instead of the full path.  The `plat_wheel` fixture changes this install name back into the full path of where `libextfunc.dylib` is located in the test data.  This keeps the platform wheel tests like before, but now they're portable.

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
 
+def _tree_libs_from_libraries(
+    libraries: Iterable[str],
+    *,
+    lib_filt_func: Callable[[str], bool],
+    copy_filt_func: Callable[[str], bool],
+    executable_path: Optional[str] = None,
+    ignore_missing: bool = False,
+) -> Dict[str, Dict[str, str]]:
+    """ Return an analysis of the dependencies of ``libraries``.
```
matthew-brett, line 255:
```suggestion
    """ Return an analysis of the dependencies of `libraries`.
```

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+        should inspect the file or False otherwise.
+        If `filt_func` filters a library it will will not further analyze any
+        of that library's dependencies.
+    copy_filt_func : None or callable
+        Called on each library name detected as a dependency; copy
+        where ``copy_filt_func(libname)`` is True, don't copy otherwise.
+    executable_path : None or str, optional
+        If not None, an alternative path to use for resolving
+        `@executable_path`.
+    ignore_missing : bool, default=False
```
matthew-brett, line 270:
```suggestion
    ignore_missing : bool, default=False, optional
```

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+    *,
+    lib_filt_func: Callable[[str], bool],
+    copy_filt_func: Callable[[str], bool],
+    executable_path: Optional[str] = None,
+    ignore_missing: bool = False,
+) -> Dict[str, Dict[str, str]]:
+    """ Return an analysis of the dependencies of ``libraries``.
+
+    libraries : iterable of str
+        The paths to the libraries to find dependencies of.
```
matthew-brett, line 258: Some comment here that all subsequent arguments are keyword-only?

matthew-brett: See: https://github.com/numpy/numpydoc/issues/330

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+    lib_filt_func : callable
+        A callable which accepts filename as argument and returns True if we
+        should inspect the file or False otherwise.
+        If `filt_func` filters a library it will will not further analyze any
+        of that library's dependencies.
+    copy_filt_func : None or callable
```
matthew-brett, line 264: Comment on what `None` means?  Should it be the default?

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+
+def tree_libs_from_directory(
+    start_path: str,
+    *,
+    lib_filt_func: Callable[[str], bool] = _filter_system_libs,
+    copy_filt_func: Callable[[str], bool] = lambda path: True,
+    executable_path: Optional[str] = None,
+    ignore_missing: bool = False,
+) -> Dict[Text, Dict[Text, Text]]:
+    """ Return an analysis of the libraries in the directory of ``start_path``.
```
matthew-brett, line 329:
```suggestion
    """ Return an analysis of the libraries in the directory of `start_path`.
```

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+
+    start_path : iterable of str
+        Root path of tree to search for libraries depending on other libraries.
+    lib_filt_func : callable, optional
```
matthew-brett, line 333: Comment as above about keyword only arguments.

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+        of that library's dependencies.
+        Defaults to inspecting all files except for system libraries.
+    copy_filt_func : callable, optional
+        Called on each library name detected as a dependency; copy
+        where ``copy_filt_func(libname)`` is True, don't copy otherwise.
+        Defaults to copying all detected dependencies.
+    executable_path : None or str, optional
+        If not None, an alternative path to use for resolving
+        `@executable_path`.
+    ignore_missing : bool, default=False
```
matthew-brett, line 346:
```suggestion
    ignore_missing : bool, default=False, optional
```

delocate/libsana.py
```diff
@@ -561,9 +699,9 @@ def wheel_libs(
     wheel_fname : str
         Filename of wheel
     filt_func : None or callable, optional
-        If None, inspect all files for library dependencies. If callable,
-        accepts filename as argument, returns True if we should inspect the
-        file, False otherwise.
+        If None, inspect all non-system files for library dependencies.
+        If callable, accepts filename as argument, returns True if we should
+        inspect the file, False otherwise.
```
matthew-brett, line 704:
```suggestion
        inspect the file, False otherwise.
    ignore_missing : bool, default=False, optional
        If False, raise error for missing dependencies.  If True, continue without error.
```

delocate/libsana.py
```diff
@@ -574,10 +712,19 @@ def wheel_libs(
         is (key, value) of (``depending_lib_path``, ``install_name``).  Again,
         ``depending_lib_path`` is library relative to wheel root path, if
         within wheel tree.
+
+    Raises
+    ------
+    DelocationError
+        When dependencies can not be located and ``ignore_missing`` is False.
```
matthew-brett, line 719:
```suggestion
        When dependencies can not be located and `ignore_missing` is False.
```

matthew-brett (commented):
Thanks - just some documentation suggestions.

delocate/libsana.py
HexDecimal: The style guide is slightly vague here:

> If the default value would not be used as a value, optional is preferred.

I take this to mean that if the default value is not something you'd ever explicitly pass to the function, then use `optional` instead of `default`.  The style guide has no examples of both `default` and `optional` being used for the same parameter.  So you use `optional` when you want to ignore or hide the default value in the function description.

delocate/libsana.py
HexDecimal: This was a slip up.  This function parameter does not allow None.

delocate/libsana.py
HexDecimal: I've added the `keyword-only` attribute to the relevant parameters.

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+        should inspect the file or False otherwise.
+        If `filt_func` filters a library it will will not further analyze any
+        of that library's dependencies.
+    copy_filt_func : callable, keyword-only
+        Called on each library name detected as a dependency; copy
+        where ``copy_filt_func(libname)`` is True, don't copy otherwise.
+    executable_path : None or str, optional, keyword-only
+        If not None, an alternative path to use for resolving
+        `@executable_path`.
+    ignore_missing : bool, default=False, keyword-only
```
matthew-brett, line 270: I know this is redundant, if the reader is reading carefully, but still ...

```suggestion
    ignore_missing : bool, default=False, optional, keyword-only
```

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+        of that library's dependencies.
+        Defaults to inspecting all files except for system libraries.
+    copy_filt_func : callable, optional, keyword-only
+        Called on each library name detected as a dependency; copy
+        where ``copy_filt_func(libname)`` is True, don't copy otherwise.
+        Defaults to copying all detected dependencies.
+    executable_path : None or str, optional, keyword-only
+        If not None, an alternative path to use for resolving
+        `@executable_path`.
+    ignore_missing : bool, default=False, keyword-only
```
matthew-brett, line 346:
```suggestion
    ignore_missing : bool, default=False, optional, keyword-only
```

delocate/libsana.py
```diff
@@ -561,9 +699,11 @@ def wheel_libs(
     wheel_fname : str
         Filename of wheel
     filt_func : None or callable, optional
-        If None, inspect all files for library dependencies. If callable,
-        accepts filename as argument, returns True if we should inspect the
-        file, False otherwise.
+        If None, inspect all non-system files for library dependencies.
+        If callable, accepts filename as argument, returns True if we should
+        inspect the file, False otherwise.
+    ignore_missing : bool, default=False, keyword-only
```
matthew-brett, line 705:
```suggestion
    ignore_missing : bool, default=False, optional, keyword-only
```

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+def _tree_libs_from_libraries(
+    libraries: Iterable[str],
+    *,
+    lib_filt_func: Callable[[str], bool],
+    copy_filt_func: Callable[[str], bool],
+    executable_path: Optional[str] = None,
+    ignore_missing: bool = False,
+) -> Dict[str, Dict[str, str]]:
+    """ Return an analysis of the dependencies of `libraries`.
+
```
matthew-brett, line 256:
```suggestion

    Parameters
    ----------
```

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+                missing_libs = True
+                continue
+            if copy_filt_func and not copy_filt_func(depending_path):
+                continue
+            lib_dict.setdefault(depending_path, {})
+            lib_dict[depending_path][library_path] = install_name
+
+    if missing_libs and not ignore_missing:
+        # Details of missing libraries would have already reported by
+        # get_dependencies.
```
matthew-brett, line 313:
```suggestion
        # get_dependencies will already have logged details of missing 
        # libraries.
```

delocate/libsana.py
```diff
@@ -243,6 +244,140 @@ def walk_directory(
...
+def tree_libs_from_directory(
+    start_path: str,
+    *,
+    lib_filt_func: Callable[[str], bool] = _filter_system_libs,
+    copy_filt_func: Callable[[str], bool] = lambda path: True,
+    executable_path: Optional[str] = None,
+    ignore_missing: bool = False,
+) -> Dict[Text, Dict[Text, Text]]:
+    """ Return an analysis of the libraries in the directory of `start_path`.
+
```
matthew-brett, line 330:
```suggestion

    Parameters
    ----------
```

delocate/tests/test_delocating.py
```diff
@@ -14,7 +14,7 @@
 from ..delocating import (DelocationError, delocate_tree_libs, copy_recurse,
                           delocate_path, check_archs, bads_report,
                           filter_system_libs)
-from ..libsana import (tree_libs, search_environment_for_lib)
+from ..libsana import (tree_libs_from_directory, search_environment_for_lib)
```
matthew-brett, line 17: Correct me if I am wrong - but I think we are no longer testing `tree_libs`.  I know we've deprecated it, but it would be good to keep the tests until we have removed it.

delocate/cmd/delocate_listdeps.py
```diff
@@ -38,10 +38,10 @@ def main():
         else:
             indent = ''
         if isdir(path):
-            lib_dict = tree_libs(path)
+            lib_dict = tree_libs_from_directory(path, ignore_missing=True)
```
matthew-brett, line 41: How about adding an `ignore-missing` optional command line parameter, defaulting to True?

delocate/cmd/delocate_listdeps.py
```diff
@@ -38,10 +38,10 @@ def main():
...
             lib_dict = stripped_lib_dict(lib_dict, realpath(getcwd()) + psep)
         else:
-            lib_dict = wheel_libs(path)
+            lib_dict = wheel_libs(path, ignore_missing=True)
```
matthew-brett, line 44: Do we have a test for `ignore_missing` and `wheel_libs`?

matthew-brett (commented):
Just a few suggestions and comments.   Looks like a nice refactor.

delocate/cmd/delocate_listdeps.py
HexDecimal: Silently ignoring missing dependencies by default would be ***really bad***.  I'd be okay with adding a command argument for this but I can't really think about adding features until the fixes from #120 are merged.

HexDecimal: I guess what I'd like is for `ignore_missing` to default to False here, but can then be enabled from the command line.  Right now I'm forced to inherit the default behavior.

delocate/cmd/delocate_listdeps.py
HexDecimal: `ignore_missing` is passed directly to `tree_libs_from_directory`, and there are tests for it on that function.

If you're worried about gaps in testing then coverage testing could be setup to check for any.

delocate/cmd/delocate_listdeps.py
HexDecimal: I'd like to do this in a separate PR if you don't mind.

matthew-brett: But - do I understand correctly that the only difference will be that, with True, the command will error with an error message, whereas, with False, it will pass without error, but will nevertheless show the missing dependencies?  In which case, we could add the parameter, default to True, and warn about the default changing to False.   Another PR is fine if you prefer.

delocate/cmd/delocate_listdeps.py
matthew-brett: Right - but it is still possible for `wheel_libs` to fail to pass the parameter correctly.

Ouch, that is embarrassing that I didn't set up coverage.   I can do that.

matthew-brett (commented):
Thanks - just a little more discussion.

delocate/cmd/delocate_listdeps.py
HexDecimal: Since I can't test locally, coverage info would have to be displayed on an online site like `codecov.io`.  I can set it up myself if you have trouble with it.

delocate/cmd/delocate_listdeps.py
HexDecimal: The current behavior is this:
`ignore_missing=True`: Suppress errors from missing dependencies, missing dependencies are logged instead of raised.
`ignore_missing=False`: Raise errors from missing dependencies.

delocate/tests/test_delocating.py
HexDecimal: Done, I've now parametrized the test.  `test_delocate_tree_libs` now tests both functions with the same test.

delocate/cmd/delocate_listdeps.py
HexDecimal: Of course this means that `ignore_missing=False` only shows one missing dependency while `ignore_missing=True` shows all of them.  There are other ways of handling this such as exiting on failure if any errors have been logged, which might be more desirable from the CLI.

`ignore_missing=False` is best for when the delocate is used as a library, or as part of tests.

delocate/cmd/delocate_listdeps.py
HexDecimal: > Right - but it is still possible for wheel_libs to fail to pass the parameter correctly.

You could technically remove `ignore_missing=ignore_missing` from the `wheel_libs` function and there's no regression test for that yet.  I'll need to figure out an easier way to set up wheels with missing dependencies.

HexDecimal: I've finally added tests for `wheel_libs`'s `ignore_missing` parameter.

delocate/cmd/delocate_listdeps.py
matthew-brett: I think we don't currently have a way of getting a full list of dependencies, when one of the dependencies is missing (unless it is the last checked).  Is that right?   The ideal would be to be able to get a full list, and error with a list of the missing dependencies, if any are missing.

HexDecimal: They can all get logged, but the functions don't return information on missing libraries since that interferes with the functions that delocate the wheels.

HexDecimal: I can try to join the exception information to include everything that was missing.  That's technically possible, and might even be relatively easy to do in Python, but it's beyond the scope of this PR.

matthew-brett:
Thanks - in it goes.

PR #122, opened 2021-09-17
Support public code coverage statistics.

HexDecimal:
Added the code coverage tools to the workflow.  There's an option to set up a YAML file but it isn't required and I haven't done it.

[The public site is here.](https://app.codecov.io/gh/matthew-brett/delocate/branch/codecov)  It already shows the coverage of my PR.  I've added a badge to the readme.

I don't really know how to handle coverage correctly without doing a development install.  I now run two tests, one for code coverage and one that's isolated from the repository.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,3 +1,4 @@
 # Requirements for tests
 wheel
 pytest
+pytest-cov
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,4 +1,3 @@
 # Requirements for tests
 wheel
 pytest
-pytest-cov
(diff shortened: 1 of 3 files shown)


test-requirements.txt
```diff
@@ -1,3 +1,4 @@
 # Requirements for tests
 wheel
 pytest
+pytest-cov
```
matthew-brett, line 4: Well - this is only so for the CI tests - right?  Can't we just install that in the CI?

HexDecimal: I've moved it to the workflow script.

matthew-brett:
Thanks for doing that.

PR #129, opened 2021-10-11
Add Black and isort formatters.

HexDecimal:
I've added Black and isort to the workflows and formatted the code with them.  When this is merged I'll no longer have to micromanage formatting and import order.

I wanted this to just be the reformatting and the config files but I had to replace some lambdas to fix warnings with Flake8.  I put off other changes I could have made so that this PR can be simpler.

I've added a `.git-blame-ignore-revs` file which can be used to ignore Black's sweepings changes when checking blames.

I've also uploaded my current VSCode workspace settings.  Which enforces an 80-column ruler and automatically formats scripts on saving.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,4 +1,6 @@
-# A common convention for ignoring commits with automated formatting.
+# File that should be passed as an argument to `git blame --ignore-revs-file`
+# See: https://www.git-scm.com/docs/git-blame
+# Name `.git-blame-ignore-revs` is a common convention, not an official Git standard.
 # To ignore anything new add the commit hash and the first line of the commit
 # message.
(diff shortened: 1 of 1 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,6 @@
+# A common convention for ignoring commits with automated formatting.
```
matthew-brett, line 1:
```suggestion
# File that will be passed as argument to `git blame --ignore-revs-file`
# Name `.git-blame-ignore-revs is a convention.
```

matthew-brett: Right - it is not automatic, I see that it could be intepreted that way.  I will edit the suggestion again.

PR #130, opened 2021-10-11
Update `parse_install_name` to the latest format.

### LINKED ISSUE #78: Can't delocate wheel with AttributeError after a regex returns no match in `parse_install_name`.
I'm running delocate on [github actions](https://github.com/matham/ffpyplayer/runs/814320068?check_suite_focus=true) on a osx build and it errors out with the following error:

```
  File "/Library/Frameworks/Python.framework/Versions/3.6/bin/delocate-wheel", line 8, in <module>
    sys.exit(main())
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/cmd/delocate_wheel.py", line 71, in main
    check_verbose=opts.verbose)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/delocating.py", line 377, in delocate_wheel
    lib_filt_func, copy_filt_func)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/delocating.py", line 289, in delocate_path
    return copy_recurse(lib_path, copy_filt_func, copied)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/delocating.py", line 145, in copy_recurse
    _copy_required(lib_path, copy_filt_func, copied_libs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/delocating.py", line 201, in _copy_required
    lib_dict = tree_libs(lib_path)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/libsana.py", line 60, in tree_libs
    for install_name in get_install_names(depending_libpath):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/tools.py", line 219, in get_install_names
    names = tuple(parse_install_name(line)[0] for line in lines[1:])
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/tools.py", line 219, in <genexpr>
    names = tuple(parse_install_name(line)[0] for line in lines[1:])
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/delocate/tools.py", line 151, in parse_install_name
    return IN_RE.match(line).groups()
AttributeError: 'NoneType' object has no attribute 'groups'
```

The command I'm running is `delocate-wheel dist/*.whl`, and the result of `ls dist/` prints `ffpyplayer-4.3.2.dev0-cp36-cp36m-macosx_10_9_x86_64.whl` indicating that the file is there.

HexDecimal:
It's likely that `otool` changed its output since #78 and `parse_install_name` needs to be updated again.

So far this changes the function so that it outputs the input string when it fails, which will make this easier to fix in the future.  I also added tests for macOS 11 and Xcode 13 but they didn't fail.

Refactors how otool install names are parsed which fixes #78.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -80,10 +80,6 @@ jobs:
       - uses: maxim-lobanov/setup-xcode@v1
         with:
           xcode-version: ${{ matrix.xcode }}
-      - name: Workaround ARM64 issues
-        # https://github.com/actions/virtual-environments/issues/2557
-        run: |
-          sudo rm -Rf /Library/Developer/CommandLineTools/SDKs/*
       - uses: actions/setup-python@v2
         with:
           python-version: "3.x"
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -67,11 +67,19 @@ jobs:
 
   isolated_tests:
     needs: tests
-    runs-on: macos-10.15
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - os: "macos-10.15"
+            xcode: "12.4"
+          - os: "macos-11"
+            xcode: "13.0"
     steps:
       - uses: maxim-lobanov/setup-xcode@v1
         with:
-          xcode-version: "12.4"
+          xcode-version: ${{ matrix.xcode }}
       - name: Workaround ARM64 issues
         # https://github.com/actions/virtual-environments/issues/2557
         run: |
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -83,6 +83,8 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: "3.x"
+      - name: Check otool version
+        run: otool --version
       - name: Download wheel
         uses: actions/download-artifact@v2
         with:
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -41,7 +41,7 @@ jobs:
           if-no-files-found: error
       - name: Install test dependencies
         run: |
-          pip install -r test-requirements.txt pytest-cov
+          pip install -r test-requirements.txt
       - name: Install delocate
         run: |
           python setup.py develop bdist_wheel
@@ -55,7 +55,7 @@ jobs:
           if-no-files-found: error
       - name: Run tests
         run: |
-          pytest --pyargs delocate --log-level DEBUG --cov-config=.coveragerc --cov=delocate --doctest-modules
+          pytest
       - name: Collect code coverage data
         run: |
           coverage xml
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -281,7 +281,7 @@ def _parse_otool_listing(stdout: str) -> Dict[str, List[str]]:
     return out
 
 
-def _ignore_architectures(input: Dict[str, T]) -> T:
+def _check_ignore_archs(input: Dict[str, T]) -> T:
     """Merge architecture outputs for functions which don't support multiple.
 
     This is used to maintain backward compatibility inside of functions which
@@ -308,13 +308,13 @@ def _ignore_architectures(input: Dict[str, T]) -> T:
     Examples
     --------
     >>> values = {"a": 10, "b": 10}
-    >>> _ignore_architectures(values)
+    >>> _check_ignore_archs(values)
     10
     >>> values
     {'a': 10, 'b': 10}
-    >>> _ignore_architectures({"": ["1", "2", "2"]})
+    >>> _check_ignore_archs({"": ["1", "2", "2"]})
     ['1', '2', '2']
-    >>> _ignore_architectures({"a": "1", "b": "not 1"})
+    >>> _check_ignore_archs({"a": "1", "b": "not 1"})
     Traceback (most recent call last):
         ...
     delocate.tools.InstallNameError: ...
@@ -442,7 +442,7 @@ def get_install_names(filename: str) -> Tuple[str, ...]:
     if not _line0_says_object(otool.stdout or otool.stderr, filename):
         return ()
     install_id = get_install_id(filename)
-    names_data = _ignore_architectures(_parse_otool_install_names(otool.stdout))
+    names_data = _check_ignore_archs(_parse_otool_install_names(otool.stdout))
     names = [name for name, _, _ in names_data]
     if install_id:  # Remove redundant install id from the install names.
         if names[0] != install_id:
@@ -471,7 +471,7 @@ def get_install_id(filename: str) -> Optional[str]:
     install_ids = _get_install_ids(filename)
     if not install_ids:
         return None  # No install ids or nothing returned.
-    return _ignore_architectures(install_ids)
+    return _check_ignore_archs(install_ids)
 
 
 def _get_install_ids(filename: str) -> Dict[str, str]:
@@ -658,7 +658,7 @@ def get_rpaths(filename: str) -> Tuple[str, ...]:
     )
     if not _line0_says_object(otool.stdout or otool.stderr, filename):
         return ()
-    rpaths = _ignore_architectures(_parse_otool_rpaths(otool.stdout))
+    rpaths = _check_ignore_archs(_parse_otool_rpaths(otool.stdout))
     return tuple(rpaths)
 
 
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -319,8 +319,8 @@ def _ignore_architectures(input: Dict[str, T]) -> T:
         ...
     delocate.tools.InstallNameError: ...
     """
-    first = next(iter(input.values()))  # Get any one value non-destructively.
-    if any(first != others for others in input.values()):
+    first, *rest = input.values()
+    if any(first != others for others in rest):
         raise InstallNameError(
             "This function does not support separate values per-architecture:"
             f" {input}"
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -408,15 +408,15 @@ def get_install_id(filename: str) -> Optional[str]:
         install id of library `filename`, or None if no install id
     """
     install_ids = _get_install_ids(filename)
-    if all(not install_id for install_id in install_ids.values()):
+    unique_ids = set(install_ids.values())
+    if not unique_ids:
         return None  # No install ids or nothing returned.
-    my_id = list(install_ids.values())[0]
-    if any(install_id != my_id for install_id in install_ids.values()):
+    if len(unique_ids) > 1:
         raise InstallNameError(
             "This function does not support separate install ids"
             f" per-architecture: {install_ids}"
         )
-    return my_id  # Only one install name.
+    return unique_ids.pop()
 
 
 def _get_install_ids(filename: str) -> Dict[str, str]:
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -18,8 +18,11 @@ from typing import (
     Sequence,
     Set,
     Tuple,
+    TypeVar,
     Union,
 )
+
+T = TypeVar("T")
 
 
 class InstallNameError(Exception):
@@ -278,6 +281,53 @@ def _parse_otool_listing(stdout: str) -> Dict[str, List[str]]:
     return out
 
 
+def _ignore_architectures(input: Dict[str, T]) -> T:
+    """Merge architecture outputs for functions which don't support multiple.
+
+    This is used to maintain backward compatibility inside of functions which
+    never supported multiple architectures.  You should not call this function
+    from new functions.
+
+    Parameters
+    ----------
+    input : dict of T
+        A dict similar to the return value of :func:`_parse_otool_listing`.
+        Must be non-empty.
+
+    Returns
+    -------
+    out : T
+        One of the values from ``input``.
+        Multiple architectures combined into a single output where possible.
+
+    Raises
+    ------
+    InstallNameError
+        If ``input`` has different values per-architecture.
+
+    Examples
+    --------
+    >>> values = {"a": 10, "b": 10}
+    >>> _ignore_architectures(values)
+    10
+    >>> values
+    {'a': 10, 'b': 10}
+    >>> _ignore_architectures({"": ["1", "2", "2"]})
+    ['1', '2', '2']
+    >>> _ignore_architectures({"a": "1", "b": "not 1"})
+    Traceback (most recent call last):
+        ...
+    delocate.tools.InstallNameError: ...
+    """
+    first = next(iter(input.values()))  # Get any one value non-destructively.
+    if any(first != others for others in input.values()):
+        raise InstallNameError(
+            "This function does not support separate values per-architecture:"
+            f" {input}"
+        )
+    return first
+
+
 def _parse_otool_install_names(
     stdout: str,
 ) -> Dict[str, List[Tuple[str, str, str]]]:
@@ -391,19 +441,13 @@ def get_install_names(filename: str) -> Tuple[str, ...]:
     )
     if not _line0_says_object(otool.stdout or otool.stderr, filename):
         return ()
-    install_ids = _get_install_ids(filename)
-    all_names = []
-    for arch, libraries in _parse_otool_install_names(otool.stdout).items():
-        names_for_arch = [name for name, _, _ in libraries]
-        if arch in install_ids:
-            # Remove redundant install id from the install names.
-            assert names_for_arch[0] == install_ids[arch]
-            names_for_arch = names_for_arch[1:]
-        for name in names_for_arch:
-            if name not in all_names:  # Avoid duplicates and preserve order.
-                all_names.append(name)
-
-    return tuple(all_names)
+    install_id = get_install_id(filename)
+    names_data = _ignore_architectures(_parse_otool_install_names(otool.stdout))
+    names = [name for name, _, _ in names_data]
+    if install_id:  # Remove redundant install id from the install names.
+        assert names[0] == install_id
+        names = names[1:]
+    return tuple(names)
 
 
 def get_install_id(filename: str) -> Optional[str]:
@@ -422,15 +466,9 @@ def get_install_id(filename: str) -> Optional[str]:
         install id of library `filename`, or None if no install id
     """
     install_ids = _get_install_ids(filename)
-    unique_ids = set(install_ids.values())
-    if not unique_ids:
+    if not install_ids:
         return None  # No install ids or nothing returned.
-    if len(unique_ids) > 1:
-        raise InstallNameError(
-            "This function does not support separate install ids"
-            f" per-architecture: {install_ids}"
-        )
-    return unique_ids.pop()
+    return _ignore_architectures(install_ids)
 
 
 def _get_install_ids(filename: str) -> Dict[str, str]:
@@ -612,14 +650,8 @@ def get_rpaths(filename: str) -> Tuple[str, ...]:
     )
     if not _line0_says_object(otool.stdout or otool.stderr, filename):
         return ()
-    all_rpaths = _parse_otool_rpaths(otool.stdout)
-    first = list(all_rpaths.values())[0]
-    if any(first != rpaths for rpaths in all_rpaths.values()):
-        raise InstallNameError(
-            "This function does not support separate rpaths per-architecture:"
-            f" {all_rpaths}"
-        )
-    return tuple(first)
+    rpaths = _ignore_architectures(_parse_otool_rpaths(otool.stdout))
+    return tuple(rpaths)
 
 
 def get_environment_variable_paths():
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -243,6 +243,17 @@ def _parse_otool_listing(stdout: str) -> Dict[str, List[str]]:
     ... example.so (architecture arm64):
     ... """)
     {'x86_64': [], 'arm64': []}
+    >>> _parse_otool_listing("")
+    Traceback (most recent call last):
+        ...
+    RuntimeError: Missing file/architecture header:...
+    >>> _parse_otool_listing("""
+    ... example.so (architecture arm64):
+    ... example.so (architecture arm64):
+    ... """)
+    Traceback (most recent call last):
+        ...
+    RuntimeError: Input has duplicate architectures for ...
     '''
     stdout = stdout.strip()
     out: Dict[str, List[str]] = {}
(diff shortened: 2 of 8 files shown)


delocate/tools.py
```diff
@@ -272,14 +407,48 @@ def get_install_id(filename):
...
+        stderr=subprocess.PIPE,
+        universal_newlines=True,
+    )
+    if not _line0_says_object(otool.stdout or otool.stderr, filename):
+        return {}
+    out = {}
+    for arch, my_id_list in _parse_otool_listing(otool.stdout).items():
+        if not my_id_list:
+            continue
+        (out[arch],) = my_id_list  # List should always have 0 or 1 items.
```
matthew-brett, line 450: How about:

```suggestion
        if my_id_list:
            out[arch] = my_id_list[0]  # List should always have 0 or 1 items.
```

delocate/tools.py
```diff
@@ -363,25 +590,22 @@ def get_rpaths(filename):
...
-        if line != "cmd LC_RPATH":
-            continue
-        cmdsize, path = lines[line_no : line_no + 2]
-        assert cmdsize.startswith("cmdsize ")
-        paths.append(RPATH_RE.match(path).groups()[0])
-        line_no += 2
-    return tuple(paths)
+    all_rpaths = _parse_otool_rpaths(otool.stdout)
+    first = list(all_rpaths.values())[0]
+    if any(first != rpaths for rpaths in all_rpaths.values()):
```
matthew-brett, line 603: This confused me a little - maybe a comment would help.   You are checking that the `rpaths` are all the same, for each architecture, if there are multiple architectures, and erroring out otherwise.   Maybe:

```suggestion
    # Check rpaths are the same for all architectures.
    rpaths_vals = list(all_rpaths.values())
    if (len(rpaths_vals) > 1 and 
        any(rpaths_vals[0] != rpaths for rpaths in rpaths_vals[1:])):
```

delocate/tools.py
```diff
@@ -272,14 +407,48 @@ def get_install_id(filename):
...
+    install_ids = _get_install_ids(filename)
+    if all(not install_id for install_id in install_ids.values()):
+        return None  # No install ids or nothing returned.
+    my_id = list(install_ids.values())[0]
+    if any(install_id != my_id for install_id in install_ids.values()):
+        raise InstallNameError(
+            "This function does not support separate install ids"
+            f" per-architecture: {install_ids}"
+        )
+    return my_id  # Only one install name.
```
matthew-brett, line 419:
```suggestion
    install_ids = _get_install_ids(filename)
    unique_ids = set(install_ids.values())
    if not unique_ids:
        return None  # No install ids or nothing returned.
    if len(unique_ids) > 1:
        raise InstallNameError(
            "This function does not support separate install ids"
            f" per-architecture: {install_ids}"
        )
    return unique_ids.pop()
```

delocate/tools.py
```diff
@@ -163,9 +173,136 @@ def parse_install_name(line):
...
+    Returns
+    -------
+    otool_out : dict of list of str
+        The dictionary key is the architecture, e.g. 'x86_64'.
+        If no architecture is parsed then the key is ''.
+        The values are a list of strings associated with the key.
+
+    Examples
+    --------
+    >>> _parse_otool_listing("""
```
matthew-brett, line 227: Convert these to tests?

delocate/tools.py
```diff
@@ -163,9 +173,136 @@ def parse_install_name(line):
...
+    ... """)
+    {'x86_64': [], 'arm64': []}
+    '''
+    stdout = stdout.strip()
+    out: Dict[str, List[str]] = {}
+    lines = stdout.split("\n")
+    while lines:
+        # Detect and parse the name/arch header line.
+        match_arch = _OTOOL_ARCHITECTURE_RE.match(lines.pop(0))
+        assert match_arch, f"Missing file/architecture header:\n{stdout}"
```
matthew-brett, line 253: Asserts to errors (asserts will be optimized away by Python -0 flags)?

delocate/tools.py
```diff
@@ -246,18 +369,30 @@ def get_install_names(filename):
...
-
-
-def get_install_id(filename):
+    install_ids = _get_install_ids(filename)
+    all_names = []
+    for arch, libraries in _parse_otool_install_names(otool.stdout).items():
+        names_for_arch = [name for name, _, _ in libraries]
+        if arch in install_ids:
+            # Remove redundant install id from the install names.
+            assert names_for_arch[0] == install_ids[arch]
```
matthew-brett, line 386: Assert to error (can be optimized away)?

delocate/tools.py
```diff
@@ -246,18 +369,30 @@ def get_install_names(filename):
...
-    install_id = get_install_id(filename)
-    if install_id is not None:
-        assert names[0] == install_id
-        return names[1:]
-    return names
-
-
-def get_install_id(filename):
+    install_ids = _get_install_ids(filename)
+    all_names = []
```
matthew-brett, line 381: Comment here to say that we're collecting all libraries we depend on, for any architecture.

(Is that always the right thing to do)?

matthew-brett (commented):
Some thoughts and suggestions.  Still trying to understand the multi-arch thing here.

.github/workflows/testing.yml
```diff
@@ -55,7 +55,7 @@ jobs:
           if-no-files-found: error
       - name: Run tests
         run: |
-          pytest --pyargs delocate --log-level DEBUG --cov-config=.coveragerc --cov=delocate --doctest-modules
+          pytest
```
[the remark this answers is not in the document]
HexDecimal: This test was always local since `--pyargs delocate` will import the local pacakge when run from the project directory.

`--pyargs delocate` becomes `testpaths` in the ` pyproject.toml` settings.  All of these settings were moved to that file.

test-requirements.txt
```diff
@@ -1,3 +1,4 @@
 # Requirements for tests
 wheel
 pytest
+pytest-cov
```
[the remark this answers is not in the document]
HexDecimal: I guess that was a habit of mine.  I like to store the coverage config options in the config files which means you need to include pytest-cov even when it's disabled.  You also get a small coverage report when you manually run tests without having to add any extra options this way.

I think the dependency is small enough that it's okay here.

matthew-brett:
Yes - but my memory was that the original tests deliberately moved out of
the source directory in order to catch the installed package.  Maybe I
forgot to do that, this time.

delocate/tools.py
HexDecimal: This is a doctest.  Examples shown in function docstrings are already checked by pytest.

The one for `_OTOOL_ARCHITECTURE_RE` isn't checked, but the functions which are using it do have tests.  It's a private regular expression which isn't exposed so I don't think it's behavior has to be locked in.

delocate/tools.py
HexDecimal: A lot of these asserts are sanity checks.  I think I can turn most of these into `RuntimeError`.

HexDecimal:
> Yes - but my memory was that the original tests deliberately moved out of
the source directory in order to catch the installed package.  Maybe I
forgot to do that, this time.

The GitHub workflows have their own isolated tests, which I know test the installed package instead of any local files because they only have the wheel to work with.

delocate/tools.py
HexDecimal: This code `(out[arch],) = my_id_list` is actually checking that the list only has one item, raising a TypeError if it has any other number of items.

`out[arch] = my_id_list[0]` would silently discard extra items, only checking that the list isn't empty.

It's possible that this is an unnecessary sanity check.  I'm not sure what to expect in the future.

delocate/tools.py
```diff
@@ -345,10 +514,68 @@ def set_install_id(filename: str, install_id: str, ad_hoc_sign: bool = True):
...
-def get_rpaths(filename):
+    Returns
+    -------
+    rpaths : dict of list of str
+        Where the key is the architecture and the value is the list of paths.
+        If the library has no rpaths then the values will be an empty list.
+
+    Examples
+    --------
+    >>> _parse_otool_rpaths("""
```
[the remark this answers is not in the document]
HexDecimal: This doctest is already tested.

delocate/tools.py
HexDecimal: I'm not sure if this is correct.  I could do what I did in the other functions and throw errors if different architectures have different results.

delocate/tools.py
HexDecimal: This was the original behavior.  I don't see any way this could be False, but I'd like to keep the assert just in case.

delocate/tools.py
HexDecimal: I've refactored duplicate code into a new `_ignore_architectures` function.  It should make everything more clear.

.github/workflows/testing.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           if-no-files-found: error
       - name: Install test dependencies
         run: |
-          pip install -r test-requirements.txt pytest-cov
+          pip install -r test-requirements.txt
```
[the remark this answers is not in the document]
HexDecimal: It isn't required, but it's easier to maintain when it's in the requirements files and not in every CI script which needs it.

delocate/tools.py
HexDecimal: I've refactored the code.  It's less complicated now.

delocate/tools.py
HexDecimal: I've made these functions more strict.

delocate/tools.py
matthew-brett: :) - yes indeed - it is a doctest!  See comment further down.

delocate/tools.py
```diff
@@ -246,18 +433,24 @@ def get_install_names(filename):
...
+    )
+    if not _line0_says_object(otool.stdout or otool.stderr, filename):
         return ()
-    names = tuple(parse_install_name(line)[0] for line in lines[1:])
     install_id = get_install_id(filename)
-    if install_id is not None:
+    names_data = _ignore_architectures(_parse_otool_install_names(otool.stdout))
+    names = [name for name, _, _ in names_data]
+    if install_id:  # Remove redundant install id from the install names.
         assert names[0] == install_id
```
matthew-brett, line 448: Repeat comment - `assert` can get optimized away, so it is safer to `raise`.

delocate/tools.py
matthew-brett: OK - but then I think, make it explicit that you are doing that check, rather than relying on the reader to infer it.  Maybe something like:

How about:

```suggestion
        if my_id_list:
            if len(my_id_list) > 1:
                raise InstallNameError("List should always have 0 or 1 items.")
            out[arch] = my_id_list[0]
```

delocate/tools.py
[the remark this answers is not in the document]
matthew-brett: As above.   Yes, we test the doctests, but it is very common for others not to test them.  If the tests were only for CI this would not be an issue.

Also - doctests are the right thing for showing someone how to use a function, but, as here, they can make the docstring verbose and harder to read, if used to do the unit testing.

delocate/tools.py
```diff
@@ -163,9 +176,197 @@ def parse_install_name(line):
...
+    {'a': 10, 'b': 10}
+    >>> _ignore_architectures({"": ["1", "2", "2"]})
+    ['1', '2', '2']
+    >>> _ignore_architectures({"a": "1", "b": "not 1"})
+    Traceback (most recent call last):
+        ...
+    delocate.tools.InstallNameError: ...
+    """
+    first = next(iter(input.values()))  # Get any one value non-destructively.
+    if any(first != others for others in input.values()):
```
matthew-brett, line 323:
```suggestion
    first, *rest = input.values()
    if any(first != others for others in rest):
```

matthew-brett (commented):
Thanks - that's a nice refactor.

delocate/tools.py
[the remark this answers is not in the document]
HexDecimal: > Yes, we test the doctests, but it is very common for others not to test them. If the tests were only for CI this would not be an issue.

You have to go out of your way to skip these tests.  I've added `--doctest-modules` to the config files, so running `pytest` alone will include them.

> Also - doctests are the right thing for showing someone how to use a function, but, as here, they can make the docstring verbose and harder to read, if used to do the unit testing.

It can make the docstring a lot larger but they're easy to skip past if you don't want to read the examples.  If these examples are hard to read then it's because I didn't always include comments inline with the example.  I don't think there can be too much documentation here.

I always prefer writing doctests to unit tests.  I use unit tests when the functions are too complex to show interactively, like when they have to work with files.  Otherwise I want to combine code examples, regression tests, and coverage into one section as much as possible.

These are non-public functions so I really wanted to keep the tests local rather than in a separate script.  Even if that meant covering exceptions in the doctests.

HexDecimal:
I've replaced the asserts you mentioned with exceptions.  There isn't a real way to make an input that covers them, so maybe I should have defended keeping asserts more.

matthew-brett:
How does keeping the asserts help in the lack of coverage?

HexDecimal:
> How does keeping the asserts help in the lack of coverage?

For example a library can't have more than one install ID (per-architecture now,) so you can't create a library which covers the test that more than one value exists.  It's just a minor issue the exists with some kinds of error checks.

matthew-brett:
Yes - I do understand that it is difficult to test - but it seems to me that's unrelated to the desirability or otherwise of asserts.   Or did I miss something?

HexDecimal:
I don't think you missed anything.  Other than maybe that's it's actually pretty uncommon to put Python into optimize mode for scripts.  Asserts being enabled is typically the default.

delocate/tools.py
```diff
@@ -163,9 +176,197 @@ def parse_install_name(line):
...
+    >>> parse_install_name("/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 905.6.0)")
+    ('/usr/lib/libc++.1.dylib', '1.0.0', '905.6.0')
+    """  # noqa: E501
     line = line.strip()
-    return IN_RE.match(line).groups()
+    match = IN_RE.match(line)
+    if not match:
+        raise ValueError(f"Could not parse {line!r}")
+    libname, compat_version, current_version = match.groups()
+    return libname, compat_version, current_version
```
matthew-brett, line 190: Doesn't the `match.groups()` already assert there are three entries?

```suggestion
    return match.groups()
```

delocate/tools.py
```diff
@@ -163,9 +176,197 @@ def parse_install_name(line):
...
+
+
+_OTOOL_ARCHITECTURE_RE = re.compile(
+    r"^(?P<name>.*?)(?: \(architecture (?P<architecture>\w+)\))?:$"
+)
+"""Matches the library separater line in 'otool -L'-like outputs.
```
matthew-brett, line 196:
```suggestion
"""Matches the library separator line in 'otool -L'-like outputs.
```

delocate/tools.py
```diff
@@ -163,9 +176,197 @@ def parse_install_name(line):
...
+                f" {current_arch!r}:\n{stdout}"
+            )
+        out[current_arch] = []
+        # Collect lines until the next header or the end.
+        while lines and not lines[0].endswith(":"):
+            out[current_arch].append(lines.pop(0).strip())
+    return out
+
+
+def _ignore_architectures(input: Dict[str, T]) -> T:
```
matthew-brett, line 284: Maybe `_check_ignore_archs`?   I missed, first time round, that this function raises for not-same values for different archs.

delocate/tools.py
```diff
@@ -246,18 +433,27 @@ def get_install_names(filename):
...
-    names = tuple(parse_install_name(line)[0] for line in lines[1:])
     install_id = get_install_id(filename)
-    if install_id is not None:
-        assert names[0] == install_id
-        return names[1:]
-    return names
-
-
-def get_install_id(filename):
+    names_data = _ignore_architectures(_parse_otool_install_names(otool.stdout))
```
matthew-brett, line 445: Maybe comment here about asserting values identical for multiple archs?

delocate/tools.py
```diff
@@ -229,7 +416,7 @@ def _line0_says_object(line0, filename):
     )
 
 
-def get_install_names(filename):
+def get_install_names(filename: str) -> Tuple[str, ...]:
```
matthew-brett, line 419: Is it tested somewhere, the behavior with multiple architectures?  Sorry if I missed it.

delocate/tools.py
```diff
@@ -246,18 +433,27 @@ def get_install_names(filename):
     install_names : tuple
         tuple of install names for library `filename`
     """
-    lines = _cmd_out_err(["otool", "-L", filename])
-    if not _line0_says_object(lines[0], filename):
+    otool = subprocess.run(
```
matthew-brett, line 436: Maybe add `Raises` section in docstring for not-matching architectures?

delocate/tools.py
```diff
@@ -363,25 +650,16 @@ def get_rpaths(filename):
     rpath : tuple
         rpath paths in `filename`
     """
-    try:
-        lines = _cmd_out_err(["otool", "-l", filename])
-    except RuntimeError:
-        return ()
-    if not _line0_says_object(lines[0], filename):
+    otool = subprocess.run(
```
matthew-brett, line 653: Consider `Raises` section in docstring for not-matching architectures.

delocate/tools.py
```diff
@@ -345,10 +574,68 @@ def set_install_id(filename: str, install_id: str, ad_hoc_sign: bool = True):
...
+            cmdsize, path = lines[line_no : line_no + 2]
+            assert cmdsize.startswith("cmdsize "), "Could not parse:\n{stdout}"
+            match_rpath = RPATH_RE.match(path)
+            assert match_rpath, "Could not parse:\n{stdout}"
+            rpaths[arch].append(match_rpath["rpath"])
+            line_no += 2
+    return rpaths
+
+
+def get_rpaths(filename: str) -> Tuple[str, ...]:
```
matthew-brett, line 638: Do we have tests somewhere for multiple architecture case?  Sorry if I missed it.

matthew-brett (commented):
Some thoughts at a quick pass.

delocate/tools.py
HexDecimal: The first assert makes sure that `match` isn't `None` which casts it to a non-Optional type.  The unpacking acts as a second cast that verifies the length of the tuple.  Your suggestion will return a literal `tuple[str, ...]` type where an `tuple[str, str, str]` is expected and this will be flagged by Mypy.

I imagine that one day the static typing will become complex enough to derive the return type from the regex string, but for now I do this.

delocate/tools.py
HexDecimal: It needs to be tested with the version of macOS that returns multiple architectures from its otool.  I couldn't figure out if GitHub Actions has a runner which has that version of the tool.  So I've been relying on feedback from the people at #78.

It'd be nice to know what version of the tools changed their output and what versions of macOS I can expect these version to be on.  Otherwise the best I can do is update the workflow to run the latest available versions of macOS.

delocate/tools.py
HexDecimal: I agree that the current name does not imply that some inputs will raise errors.  I'll change it.

delocate/tools.py
HexDecimal: I might need to change the exceptions to tell the difference between a `_check_ignore_archs` failure and other issues.

delocate/tools.py
HexDecimal: Same here as `get_install_names`.

delocate/tools.py
HexDecimal: I've added `Raises` sections to multiple functions.  I'm now using `NotImplementedError` to indicate that a new function that can handle multiple multiple architectures needs to be added eventually.

delocate/tools.py
HexDecimal: This is now indicated on the functions `Raises` section.

delocate/tools.py
matthew-brett: How about using mock on `subprocess` to simulate the output for multiple architectures?

HexDecimal: That could work, but the functions which normalize the output of otool are already tested with the alternative output.  I'd rather test on the real tools but I can't even start with that without knowing when the change in output happened.

matthew-brett: That would be good - to test on the real tools - but I worry that we have a currently untested change of behavior for these two public functions - even if their subfunctions are tested.  I mean, we could change these two public functions, and give a different output, and the tests would not detect this.

HexDecimal: There isn't a regression or change to any existing functions.  That would show up on the current tests, and I haven't changed any tests to expect a different output.  This is the way I see it:

| | Old behavior| After PR
|--|---|-------
| Old otool output | Works | Works 
| New otool output | Broken according to feedback | Works according to feedback

I actually can't change their behavior and these functions will have to be deprecated and replaced because their current spec is incompatible with how architectures are handled by the new tools.  When new behavior is required to handle a library then `NotImplemendError` will be raised.  So far I found this to be rare, but I expect an issue to be raised eventually.  The corrected version of these functions would look like `_get_install_ids` where they return values for each architecture as a dictionary.

matthew-brett: Sure - no question this PR is better than the previous situation.  But - it's still true that we currently have changes here, that work for the new otool output, that we could change, and we would not detect this in the tests.  I'm happy to merge as-is, and put in some tests soonish, if you like.

HexDecimal: >  But - it's still true that we currently have changes here, that work for the new otool output, that we could change, and we would not detect this in the tests. 

I understand that, but there isn't much I can do since I don't know enough about macOS to install the updated tools on the GitHub Runners.

I've done my best in this situation.  I have the changes in output handled by two functions that can test whatever input I want to give them: `_parse_otool_listing` and `_check_ignore_archs`.

`_parse_otool_listing` interprets the output and can handle the old and new output.  With the new output it attaches the architecture to each value since that's now part of otools output.  With the old output it attaches a blank `""` to the value.  The values are the same 'list of lines' as before, but now with an optional architecture as the keys of a dictionary.

`_check_ignore_archs` wraps `_parse_otool_listing` and ensures that the old and new output give the same result.  A single item / single architecture / blank architecture has the single value returned as-is.  With multiple items the values are returned as one only if they're all equivalent, which mimics the expected result of the old tools.  If the values don't match then the calling function needs to handle the output of `_parse_otool_listing` directly so `NotImplementedError` gets raised in this case.

With this the difference in otools output can be handled and tested by these two functions using fake otool output based on feedback and the existing older output.  I have high confidence that this works how I've explained it.

HexDecimal: Maybe a tl;dr:

`_parse_otool_listing` fixes #78 by parsing both the old and new output.

`_check_ignore_archs` is an adapter for the existing functions.  It returns what the old output would've returned and is tested with both the old and new output.

This means the output is normalized by the time a caller like `get_install_names` gets to handle it.

matthew-brett: Right - but we're going round in circles a bit here - I'm sure you agree that it's currently possible to change `get_install_names` and `get_rpaths` so that the multi-arch stuff breaks, and the tests continue to pass.  The question is - would you like me to merge now and we'll do some tests later?

HexDecimal: > The question is - would you like me to merge now and we'll do some tests later?

Sorry.  Yes I'd like it merged its current state.  I can make a new issue just to track that the new tools aren't directly tested yet.  Maybe someone else will bring in with more information.

> I'm sure you agree that it's currently possible to change get_install_names and get_rpaths so that the multi-arch stuff breaks, and the tests continue to pass.

Sorry again, but maybe I'm confused by the phrasing.  These functions are already like this: broken when they'd have to return different results per-architecture, but I haven't heard anyone run into this yet whenever I ask them to install this PR.

matthew-brett:
As ever - thanks for working so quickly on these issues.

Issue #133, opened 2021-11-16
DelocationError: Already planning to copy library with same basename as: ...


sighingnow:
**Describe the bug**

Hi @matthew-brett,

It seems that `delocate` cannot process brew-installed libraries correct, e.g., if we have two binary target in a wheel package, one of it links to `/usr/local/lib/libabsl_status.2103.0.1.dylib` and another links to `/usr/local/Cellar/abseil/20210324.2_1/lib/libabsl_status.2103.0.1.dylib`, `delocate` will raise exception:

```
2021-11-16T05:00:04.8377350Z Fixing: dist/gs_lib-0.8.0-py2.py3-none-macosx_10_9_x86_64.whl
2021-11-16T05:00:04.8378500Z   File "/usr/local/bin/delocate-wheel", line 8, in <module>
2021-11-16T05:00:04.8379030Z     sys.exit(main())
2021-11-16T05:00:04.8380080Z   File "/usr/local/lib/python3.9/site-packages/delocate/cmd/delocate_wheel.py", line 91, in main
2021-11-16T05:00:04.8380810Z     copied = delocate_wheel(
2021-11-16T05:00:04.8381920Z   File "/usr/local/lib/python3.9/site-packages/delocate/delocating.py", line 598, in delocate_wheel
2021-11-16T05:00:04.8382640Z     copied_libs = delocate_path(
2021-11-16T05:00:04.8383750Z   File "/usr/local/lib/python3.9/site-packages/delocate/delocating.py", line 429, in delocate_path
2021-11-16T05:00:04.8384830Z     return delocate_tree_libs(lib_dict, lib_path, tree_path)
2021-11-16T05:00:04.8386340Z   File "/usr/local/lib/python3.9/site-packages/delocate/delocating.py", line 80, in delocate_tree_libs
2021-11-16T05:00:04.8387230Z     libraries_to_copy, libraries_to_delocate = _analyze_tree_libs(
2021-11-16T05:00:04.8388510Z   File "/usr/local/lib/python3.9/site-packages/delocate/delocating.py", line 121, in _analyze_tree_libs
2021-11-16T05:00:04.8389720Z     raise DelocationError('Already planning to copy library with '
2021-11-16T05:00:04.8390860Z delocate.delocating.DelocationError: Already planning to copy library with same basename as: libabsl_status.2103.0.1.dylib
2021-11-16T05:00:04.8516800Z make: *** [graphscope-darwin-py3] Error 1
```

[...]

PR #134, opened 2021-11-20
Avoid duplication when library has soft links and been used in the same wheels.

sighingnow:
Resolves issue #133.

I have managed to reproduce the issue by the following minimal case:

We have:

```
  liba.dylib
  links/liba.dylib, a soft link to liba.dylib
```

and

```
  libb.dylib, depends on liba.dylib via `@rpath/liba.dylib`
  links/libb.dylib, depends on links/liba.dylib and been found via `DYLD_LIBRARY_PATH`
```

During resolving dependencies, the first is found via resolving from `@rpath` by `resolve_dynamic_paths`, and the later is resolved from `DYLD_LIBRARY_PATH` by `search_environment_for_lib`, where the soft link path is returned, thus the conflicit happens.

For more details, see the minimal reproduce case in `test_tree_libs_from_directory_with_links`, and we have verified the patch in our real-world wheels that triggerred #133 before.

### DIFF
--- a/delocate/libsana.py
+++ b/delocate/libsana.py
@@ -625,7 +625,7 @@ def search_environment_for_lib(lib_path):
 
     for location in potential_library_locations:
         if os.path.exists(location):
-            return location
+            return realpath(location)
     return realpath(lib_path)
 
 
--- a/delocate/libsana.py
+++ b/delocate/libsana.py
@@ -604,7 +604,7 @@ def search_environment_for_lib(lib_path):
     Returns
     -------
     lib_path : str
-        Full path of ``basename(lib_path)``'s location, if it can be found, or
+        Real path of the first found location, if it can be found, or
         ``realpath(lib_path)`` if it cannot.
     """
     lib_basename = basename(lib_path)
@@ -625,6 +625,7 @@ def search_environment_for_lib(lib_path):
 
     for location in potential_library_locations:
         if os.path.exists(location):
+            # See GH#133 for why we return the realpath here if it can be found
             return realpath(location)
     return realpath(lib_path)
 
--- a/delocate/tests/test_libsana.py
+++ b/delocate/tests/test_libsana.py
@@ -302,7 +302,7 @@ def test_tree_libs_from_directory_with_links() -> None:
                 libb_use_link,
             ]
         )
-        # hack libb.dylib to resolving from rpath, bypass searching from env
+        # hack libb.dylib to resolve from rpath, bypass searching from env
         subprocess.check_call(
             [
                 "install_name_tool",
--- a/delocate/tests/test_libsana.py
+++ b/delocate/tests/test_libsana.py
@@ -334,7 +334,10 @@ def test_tree_libs_from_directory_with_links() -> None:
 
         # Put dir of soft link for `liba.dylib` into `DYLD_LIBRARY_PATH`
         with TempDirWithoutEnvVars("DYLD_LIBRARY_PATH"):
-            assert tree_libs_from_directory(tmpdir) != exp_dict
+            # the result should be correct normally
+            assert tree_libs_from_directory(tmpdir) == exp_dict
+
+            # the result should be correct even if there are soft links in `$DYLD_LIBRARY_PATH`
             os.environ["DYLD_LIBRARY_PATH"] = os.path.dirname(liba_link)
             assert tree_libs_from_directory(tmpdir) == exp_dict
 
--- a/delocate/tests/test_libsana.py
+++ b/delocate/tests/test_libsana.py
@@ -332,15 +332,10 @@ def test_tree_libs_from_directory_with_links() -> None:
         )
 
         # Put dir of soft link for `liba.dylib` into `DYLD_LIBRARY_PATH`
-        dyld_library_path = os.environ.get("DYLD_LIBRARY_PATH")
-        os.environ["DYLD_LIBRARY_PATH"] = os.path.dirname(liba_link)
-        try:
+        with TempDirWithoutEnvVars("DYLD_LIBRARY_PATH"):
+            assert tree_libs_from_directory(tmpdir) != exp_dict
+            os.environ["DYLD_LIBRARY_PATH"] = os.path.dirname(liba_link)
             assert tree_libs_from_directory(tmpdir) == exp_dict
-        finally:  # Restore os.environ.
-            if dyld_library_path is not None:
-                os.environ["DYLD_LIBRARY_PATH"] = dyld_library_path
-            else:
-                del os.environ["DYLD_LIBRARY_PATH"]
 
 
 def test_get_prefix_stripper():
--- a/delocate/tests/test_libsana.py
+++ b/delocate/tests/test_libsana.py
@@ -331,7 +331,7 @@ def test_tree_libs_from_directory_with_links() -> None:
             }
         )
 
-        # makes the soft link of `liba.dylib` presents in `DYLD_LIBRARY_PATH`
+        # Put dir of soft link for `liba.dylib` into `DYLD_LIBRARY_PATH`
         dyld_library_path = os.environ.get("DYLD_LIBRARY_PATH")
         os.environ["DYLD_LIBRARY_PATH"] = os.path.dirname(liba_link)
         try:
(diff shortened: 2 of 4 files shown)


sighingnow:
This pull request generates copied files in the test case rather than put it into `tests/data` to avoid affecting other test cases.

delocate/tests/test_libsana.py
```diff
@@ -251,6 +252,105 @@ def filt(fname: str) -> bool:
...
+        os.environ["DYLD_LIBRARY_PATH"] = new_dyld_library_path
+
+        # check the result
+        assert tree_libs_from_directory(tmpdir) == exp_dict
+
+        # restore os environment
+        if dyld_library_path is not None:
+            os.environ["DYLD_LIBRARY_PATH"] = dyld_library_path
+        else:
+            del os.environ["DYLD_LIBRARY_PATH"]
```
HexDecimal, line 351: If the test fails here then `DYLD_LIBRARY_PATH` won't be restored for the remaining tests.  Would you mind wrapping this in a context manager?

Also `os.path.pathsep.join` takes an iterator which wasn't given.  This might be over engineered and you'll do fine with just overwriting `DYLD_LIBRARY_PATH` without appending to it.

Is `DYLD_LIBRARY_PATH` case sensitive?

HexDecimal (changes requested):
The fix itself looks good.

You should add a note for this fix in the changelog.

I'd accept the test if it was at least using a try..finally block when working with `os.environ`.

delocate/tests/test_libsana.py
HexDecimal:
```suggestion
        # makes the soft link of `liba.dylib` presents in `DYLD_LIBRARY_PATH`
        dyld_library_path = os.environ.get("DYLD_LIBRARY_PATH")
        os.environ["DYLD_LIBRARY_PATH"] = os.path.dirname(liba_link)
        try:
            assert tree_libs_from_directory(tmpdir) == exp_dict
        finally:  # Restore os.environ.
            if dyld_library_path is not None:
                os.environ["DYLD_LIBRARY_PATH"] = dyld_library_path
            else:
                del os.environ["DYLD_LIBRARY_PATH"]
```

sighingnow: > If the test fails here then `DYLD_LIBRARY_PATH` won't be restored for the remaining tests. Would you mind wrapping this in a context manager?
> 

Thanks for the suggestion. Will add it.

> Also `os.path.pathsep.join` takes an iterator which wasn't given. This might be over engineered and you'll do fine with just overwriting `DYLD_LIBRARY_PATH` without appending to it.
> 

Yes, overwriting should be enough.

> Is `DYLD_LIBRARY_PATH` case sensitive?

At least in our codebase, yes, the environment variable name is case sensitive.

sighingnow:
> The fix itself looks good.
> 
> You should add a note for this fix in the changelog.
> 
> I'd accept the test if it was at least using a try..finally block when working with `os.environ`.

I have applied the suggested change.

sighingnow:
Hi @HexDecimal could we get this one merge? and it it is possible, could we have a new release on pypi to include this fix?

Thanks!

HexDecimal (approved):
It's good as far as I can tell.  I've noted the fix in the changelog.

HexDecimal:
@matthew-brett, would you look at the test in case I've missed anything obvious?  Or should I merge PR's without your input?

I can make a bug fix release immediately once this is merged.

matthew-brett:
Thanks both.  Yes, will have a look at this tomorrow morning, but @HexDecimal - feel free to merge without me if I'm being slow, or a merge needs doing.

delocate/libsana.py
```diff
@@ -625,7 +625,7 @@ def search_environment_for_lib(lib_path):
 
     for location in potential_library_locations:
         if os.path.exists(location):
-            return location
+            return realpath(location)
```
matthew-brett, line 628: Would you mind changing the docstring above to say that we return the realpath of the first found location, or the realpath if not found.

delocate/tests/test_libsana.py
```diff
@@ -251,6 +252,98 @@ def filt(fname: str) -> bool:
         )
 
 
+def test_tree_libs_from_directory_with_links() -> None:
+    # Test ability to walk through tree, where the same library may have
+    # soft links under different subdirectories. See also GH#133, where
+    #
```
matthew-brett, line 258:
```suggestion
```

delocate/tests/test_libsana.py
```diff
@@ -251,6 +252,98 @@ def filt(fname: str) -> bool:
...
+
+        # copy files
+        os.makedirs(pjoin(tmpdir, "links"))
+        liba_link = pjoin(tmpdir, "links", "liba.dylib")
+        libb_use_link = pjoin(tmpdir, "links", "libb.dylib")
+        rp_libb_use_link = realpath(libb_use_link)
+        os.symlink(liba, liba_link)
+        shutil.copy2(libb, libb_use_link)
+
+        # hack links/libb.dylib to depends on the softlink of liba.dylib
```
matthew-brett, line 296:
```suggestion
        # hack links/libb.dylib to depend on the softlink of liba.dylib
```

delocate/tests/test_libsana.py
```diff
@@ -251,6 +252,98 @@ def filt(fname: str) -> bool:
...
+        subprocess.check_call(
+            [
+                "install_name_tool",
+                "-change",
+                "liba.dylib",
+                liba_link,
+                libb_use_link,
+            ]
+        )
+        # hack libb.dylib to resolving from rpath, bypass searching from env
```
matthew-brett, line 306:
```suggestion
        # hack libb.dylib to resolve from rpath, bypass searching from env
```

delocate/tests/test_libsana.py
```diff
@@ -251,6 +252,98 @@ def filt(fname: str) -> bool:
...
+        exp_dict.update(
+            {
+                rp_liba: {
+                    rp_libb: "@rpath/liba.dylib",
+                    rp_libb_use_link: liba_link,
+                },
+            }
+        )
+
+        # makes the soft link of `liba.dylib` presents in `DYLD_LIBRARY_PATH`
```
matthew-brett, line 335:
```suggestion
        # Put dir of soft link for `liba.dylib` into `DYLD_LIBRARY_PATH`
```

delocate/tests/test_libsana.py
```diff
@@ -251,6 +252,98 @@ def filt(fname: str) -> bool:
...
+        dyld_library_path = os.environ.get("DYLD_LIBRARY_PATH")
+        os.environ["DYLD_LIBRARY_PATH"] = os.path.dirname(liba_link)
+        try:
+            assert tree_libs_from_directory(tmpdir) == exp_dict
+        finally:  # Restore os.environ.
+            if dyld_library_path is not None:
+                os.environ["DYLD_LIBRARY_PATH"] = dyld_library_path
+            else:
+                del os.environ["DYLD_LIBRARY_PATH"]
```
matthew-brett, line 344: Maybe add:

```
from .env_tools import TempDirWithoutEnvVars:
```

at top of file, and

```suggestion
        with TempDirWithoutEnvVars("DYLD_LIBRARY_PATH"):
            assert tree_libs_from_directory(tmpdir) != exp_dict
            os.environ["DYLD_LIBRARY_PATH"] = os.path.dirname(liba_link)
            assert tree_libs_from_directory(tmpdir) == exp_dict
```

matthew-brett (commented):
Thanks for this - just a few suggestions.

delocate/libsana.py
sighingnow: Fixed.

sighingnow:
> Thanks for this - just a few suggestions.

Thanks for reviewing, all comments have been addressed.

sighingnow:
Lint error fixed.

matthew-brett:
Thanks very much @sighingnow  for the PR, and @HexDecimal for the review.

HexDecimal:
You're welcome. I've also released a version with this fix.

sighingnow:
> You're welcome. I've also released a version with this fix.

Thanks a lot.

PR #136, opened 2021-12-31
Replicate the to-be-removed `wheel.pkginfo` module.

### LINKED ISSUE #135: Latest development main of wheel is missing pkginfo
**Describe the bug**
`wheel.pkginfo` was removed.

**To Reproduce**
Use an unreleased version of wheel.

**Expected behavior**
Not break.

**Wheels used**
The import is missing, so no need for a wheel.

 - OS version: any
 - Delocate version: any

**Additional context**

See: https://github.com/pypa/wheel/issues/433

HexDecimal:
The plans to remove the `wheel.pkginfo` module can be negotiated with the `wheel` devs, but this PR should cause less drama.

Adds `delocate.pkginfo` to read/write these files, based on the original module.

Removes code supporting `wheel<0.32.0`.  Adds code to support the unreleased versions of `wheel>0.37.1`.

Adds type hinting to the relevant functions.

Closes #135

### DIFF
--- a/Changelog
+++ b/Changelog
@@ -21,6 +21,10 @@ Releases
 
 * [Unreleased]
 
+  * Ensured support for versions of wheel beyond 0.37.1.  If you are using older
+    versions of Delcoate then you may need to to pin ``wheel==0.37.1``.
+    [#135](https://github.com/matthew-brett/delocate/issues/135)
+
 * [0.10.1] - 2021-11-23
 
   Bugfix release.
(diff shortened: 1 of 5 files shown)


matthew-brett:
Worth also copying the `test_pkginfo.py` test?

HexDecimal:
> Worth also copying the `test_pkginfo.py` test?

I thought of that, but looking at the test and the code, the test is just making sure that `email.generator` works rather than anything more specific to the function itself.

All 3 functions are covered by other tested functions which rely on them.

matthew-brett:
OK - nice - thanks for doing that so fast.

PR #138, opened 2022-01-01
Setup cross-platform testing,

HexDecimal:
I've went and marked which tests fail on which platforms.  After that I added more platforms to the testing workflow.  Strict xfail is used to make sure that the tests marked as known-to-fail are actually failing, which makes this a good way to track which functions need to be improved to be cross-platform.

I've discovered that the [unzip tool](https://github.com/matthew-brett/delocate/blob/a5ba7dcc8ccfc9aaa1411c63266bfaf87a4ff393/delocate/tools.py#L757-L774) used by Delocate does not work on Windows, since it's normally missing the unzip program.

I'm not in hurry to merge this, but I don't think this needs a full review before merging.  I'll let this PR sit for a while.


matthew-brett:
While I remember - we're using an external `unzip` instead of the `zipfile` module because of https://bugs.python.org/issue15795

HexDecimal:
I'm just pointing it out since I didn't notice it until I started marking tests.  The function itself has the link to the Python bug which itself has links to some workarounds which would work on Windows, but I don't really want to mess with actually porting Delocate.  I just want to be able to test small platform independent features without the Mac-only tests getting in the way.

matthew-brett (approved):
LGTM

PR #140, opened 2022-01-01
MRG: Sketch of mock testing for get_install_names

matthew-brett:
Use mock to simulate otool output for get_install_names

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,5 +1,5 @@
 [mypy]
-python_version = 3.6
+python_version = 3.7
 
 [mypy-wheel.*]
 ignore_missing_imports = True
--- a/delocate/tests/test_install_names.py
+++ b/delocate/tests/test_install_names.py
@@ -1,9 +1,20 @@
 """ Tests for install name utilities """
 
+import contextlib
 import os
 import shutil
 from os.path import basename, dirname, exists
 from os.path import join as pjoin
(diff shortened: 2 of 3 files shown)


HexDecimal:
Add `# noqa: E501` to the lines you close the multi-line strings with since there isn't a graceful way to shorten them.  Running black should resolve any remaining formatting issues.

matthew-brett:
Done.

matthew-brett:
I added testing for Rpaths and install IDs.  It's a bit verbose, but I think it is better than not testing, and it does give a template for Otool output difference testing.

HexDecimal:
I've just pushed a significant refactor to this PR, solving every issue I had with it.  If you approve of my changes then this is should be merged.

matthew-brett (commented):
Looks good to me.  Did you test that the tests fail when the inputs are not as expected?

HexDecimal:
I'm not sure what you mean.  The mock is more strict requiring that `example.so` be the file tested, so changing that will fail the tests.  Changing the mocked outputs will fail tests if the changed parts are relevant to the called functions.  The parametrized tests will show which `ToolArchMock` failed to pass.

matthew-brett:
OK - I spent some time randomly deleting and modifying the `ToolArchMock` entries and confirming that the test fails.

PR #141, opened 2022-01-03
Remove dependencies on wheels unstable API.

HexDecimal:
Based on feedback from https://github.com/pypa/wheel/issues/433.  There are plans on refactoring wheel which will likely break Delocate again.  This PR will prevent that from happening.

Small single-used wheel functions are now inline.  The packaging module was added which has a stable API for parsing wheel names.  Type hinting added to relevant functions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
     install_requires=[
         "machomachomangler; sys_platform == 'win32'",
         "bindepend; sys_platform == 'win32'",
-        "packaging",
+        "packaging>=20.9",
         "typing_extensions",
     ],
     package_data={
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
     install_requires=[
         "machomachomangler; sys_platform == 'win32'",
         "bindepend; sys_platform == 'win32'",
-        "wheel>=0.32.0",
+        "packaging",
         "typing_extensions",
     ],
     package_data={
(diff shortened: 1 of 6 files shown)


matthew-brett:
Thanks - looks good to me.  Would you mind merging when CI passes?

HexDecimal:
> Would you mind merging when CI passes?

TravisCI seems to timeout before it can run.  I also can't manually re-run those jobs on a PR.

setup.py
```diff
@@ -20,7 +20,7 @@
     install_requires=[
         "machomachomangler; sys_platform == 'win32'",
         "bindepend; sys_platform == 'win32'",
-        "wheel>=0.32.0",
+        "packaging",
```
mayeut, line 23:
```suggestion
        "packaging>=20.9",
```

`parse_wheel_filename` was added in 20.9: https://github.com/pypa/packaging/commit/da68002685b0a7d0027e189129d6bde35303a58b

HexDecimal:
I guess the question is should I merge this when TravisCI has timed out and what's going to be done about TravisCI in the long term since it only seems to become more unreliable?

matthew-brett:
Yes, please do merge after rebase (looks like).

And for Travis-CI - yes - we should plan to drop it as soon as possible.  One of us should go through and check, record what we are losing, when we do - I will have a chance to do that next week if you can't get to it before.

PR #146, opened 2022-01-21
Refactor calls to CLI tools.

HexDecimal:
I noticed from #144 that stderr was not being shown on errors from `subprocess.run`.  That was a thing before with the `back_tick` function so I've added this feature back.

Duplicate code has been moved into a new private function.  This function provides better error messages when a command fails, by showing the full output of the failing command.


matthew-brett (approved):
Nice refactor.  LGTM.

PR #159, opened 2022-09-04
Remove TravisCI integration.

### LINKED ISSUE #145: TravisCI timeout issues.
TravisCI has issues running lately and will timeout instead of running tests.  Github workflows are working and are up-to-date so this isn't a critical issue, but the failing status can be annoying to look at.  I'm making this issue to track any discussion and progress.

https://github.com/matthew-brett/delocate/pull/141#issuecomment-1010910183
> And for Travis-CI - yes - we should plan to drop it as soon as possible. One of us should go through and check, record what we are losing, when we do - I will have a chance to do that next week if you can't get to it before.

HexDecimal:
TravisCI used to be great, but now it is no longer capable of supporting open-source projects like this one.
In its current state this CI only exists to cause a failure status.

Coverage is already handled by new workflows.  They also support `.coveragerc`.

The current workflows integrate with multibuild, and run tests with the same log levels.

Testing on older versions of MacOS is not be possible in general due to a lack of support from Apple for older MacOS versions.

Closes #145


matthew-brett:
Agreed, time for the long march to end ...

PR #161, opened 2022-09-27
GitHub Actions updates

radarhere:
Updates several GitHub Actions, as well as switch from ubuntu-20.04 to ubuntu-latest


HexDecimal (approved):
`ubuntu-20.04` isn't deprecated and `ubuntu-latest` points to `ubuntu-20.04` instead of the actual latest version `ubuntu-22.04`.

I'd guess this is okay since the platform specifics of Linux are not as important for this project, but normally I'd recommend exact versions for `runs-on`.

HexDecimal:
I'd keep the small lint jobs on `ubuntu-latest` though.

Issue #169, opened 2022-11-22
Please describe `delocate-fuse` in the README


rgommers:
`delocate-fuse` seems like the easy to use command-line utility one needs to produce `universal2` wheels (thanks for adding it to this project Matthew!). It'd be great to have it described in the README, so it is easier to discover. Especially because I and others have been pointing people to it as the recommended way to get a `universal2` wheel from thin wheels. It seems like the CLI tool that `py2app` & co need.

mattip:
Could you make a release with this and #167 so that NumPy (and maybe others) do not have to install from git?

HexDecimal:
This is only a doc update, but I'll make a new release with #167.

mattip:
Thanks

HexDecimal:
@mattip delocate 0.10.4 has been released.

mattip:
That was quick, thanks!

PR #170, opened 2022-11-22
DOC: document delocate-fuse

matthew-brett:
Closes gh-169

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -161,6 +161,43 @@ fault by adding the ``-v`` (verbose) flag::
 
 I need to rebuild this wheel to link with dual-architecture libraries.
 
+Making dual-architecture binaries
+=================================
+
+Modern Mac wheels can be either ``arm64e`` (M1 ARM), ``x86_64`` (64-bit Intel) or
+both ("universal").
+
+Building an entire Python wheel as dual-architecture can be difficult, perhaps
+because you need to link different libraries in the two cases, or you need
+different compiler flags.
+
+One solution to this problem is to do an entire ``arm64e`` wheel build, and then
+an entire ``x86_64`` wheel build, and *fuse* the two wheels into a universal
+wheel.
+
+That is what the ``delocate-fuse`` command does.
+
+Let's say you have built an ARM and Intel wheel, called, respectively:
+
+* ``scipy-1.9.3-cp311-cp311-macosx_12_0_arm64.whl``
+* ``scipy-1.9.3-cp311-cp311-macosx_10_9_x86_64.whl``
+
+Then you could create a new fused (universal) wheel in the `tmp` subdirectory with::
+
+    delocate-fuse scipy-1.9.3-cp311-cp311-macosx_12_0_arm64.whl scipy-1.9.3-cp311-cp311-macosx_10_9_x86_64.whl -w tmp
+
+The output wheel in that case would be:
+
+* ``tmp/scipy-1.9.3-cp311-cp311-macosx_12_0_arm64.whl``
+
+You will find, using ``lipo -archs`` - that all binaries with the same name in
+each wheel are now universal (``x86_64`` and ``arm64``).
+
+To be useful, you should rename it to reflect the fact that it is now
+a universal wheel - in this case to:
+
+* ``tmp/scipy-1.9.3-cp311-cp311-macosx_12_0_universal2.whl``
+
 Troubleshooting
 ===============
 
--- a/README.rst
+++ b/README.rst
@@ -120,7 +120,13 @@ are both dual architecture binaries.  For example::
     $ lipo -info /usr/bin/python
     Architectures in the fat file: /usr/bin/python are: x86_64 arm64e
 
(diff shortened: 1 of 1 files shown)


matthew-brett:
@rgommers - any chance of a quick review?

README.rst
```diff
@@ -161,6 +161,43 @@ fault by adding the ``-v`` (verbose) flag::
 
 I need to rebuild this wheel to link with dual-architecture libraries.
 
+Making dual-architecture binaries
+=================================
+
+Modern Mac wheels can be either ``arm64e`` (M1 ARM), ``x86_64`` (64-bit Intel) or
```
rgommers, line 167: spurious "e" in `arm64` here?

README.rst
```diff
@@ -161,6 +161,43 @@ fault by adding the ``-v`` (verbose) flag::
...
+both ("universal").
```
rgommers, line 168: `universal2`

README.rst
```diff
@@ -161,6 +161,43 @@ fault by adding the ``-v`` (verbose) flag::
...
+
+Building an entire Python wheel as dual-architecture can be difficult, perhaps
+because you need to link different libraries in the two cases, or you need
+different compiler flags.
```
rgommers, line 172: maybe add "or because you build for `arm64` on native (M1) CI". May or may not want to mention Cirrus CI, which isn't very well-known yet but offers free M1 CI which works very well.

README.rst
rgommers: hmm you repeat it below, so I guess it's on purpose - but I've never seen the `e` before

README.rst
```diff
@@ -161,6 +161,43 @@ fault by adding the ``-v`` (verbose) flag::
...
+One solution to this problem is to do an entire ``arm64e`` wheel build, and then
+an entire ``x86_64`` wheel build, and *fuse* the two wheels into a universal
+wheel.
+
+That is what the ``delocate-fuse`` command does.
+
+Let's say you have built an ARM and Intel wheel, called, respectively:
+
+* ``scipy-1.9.3-cp311-cp311-macosx_12_0_arm64.whl``
+* ``scipy-1.9.3-cp311-cp311-macosx_10_9_x86_64.whl``
```
rgommers, line 183: consider replacing `scipy` with `pkgname` for a bit more genericity, and `12_0` with `11_0` because that'd be the normal tag (SciPy is special here, because its test suite triggers a kernel panic on 11.0).

README.rst
```diff
@@ -161,6 +161,43 @@ fault by adding the ``-v`` (verbose) flag::
...
+
+Then you could create a new fused (universal) wheel in the `tmp` subdirectory with::
```
rgommers, line 185: `universal2`

README.rst
```diff
@@ -161,6 +161,43 @@ fault by adding the ``-v`` (verbose) flag::
...
+
+The output wheel in that case would be:
+
+* ``tmp/scipy-1.9.3-cp311-cp311-macosx_12_0_arm64.whl``
+
+You will find, using ``lipo -archs`` - that all binaries with the same name in
+each wheel are now universal (``x86_64`` and ``arm64``).
+
+To be useful, you should rename it to reflect the fact that it is now
+a universal wheel - in this case to:
```
rgommers, line 197: Is the output wheel always the first wheel given? If it can be done out of place, that should probably be preferred. Otherwise it'd be good to say something like "Note that the first wheel is updated in-place" right below the `delocate-fuse` usage.

rgommers (commented):
Thanks for the quick response Matthew. A few comments

README.rst
matthew-brett: I believe `arm64e` refers to some extended features on ARM : see https://stackoverflow.com/questions/52624308/xcode-arm64-vs-arm64e

I found it doing `lipo -archs /usr/bin/python3` on my M2 machine.  I'm guessing that `arm64e` is valid for M1 machines also.  But yes - probably best to stick to `arm64` and maybe do a note about `arm64e`.

rgommers: Ah okay, learned something new. But yes, it's unusual (first time I've seen it) so probably better to stick to plain `arm64`.

README.rst
matthew-brett: I've used Scipy as an example a lot above, so I'll stick with Scipy here - otherwise I think it would be more confusing.  I've changed to `11_0`.

README.rst
matthew-brett: Done.

README.rst
matthew-brett: Done.

README.rst
matthew-brett: Yes, it would be good to add a flag to work out the output wheel name and use that instead.   Or provide your own output filename.

But I've noted the current behavior.

rgommers (approved):
LGTM now, thanks @matthew-brett

matthew-brett:
Thanks - in it goes.

PR #172, opened 2023-01-23
Speed-up delocate on large wheels.

colesbury:
Mach-O executables and libraries begin with a magic number. It's much faster to read the first few bytes of a file than to spawn `otool`.

For example, `delocate list-deps` currently takes about 12 minutes on the PyTorch macOS wheels (and nearly twice as long on GitHub's CI runners). With this change, `delocate list-deps` instead takes about 40 seconds.

### DIFF
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -162,7 +162,7 @@ MACHO_MAGIC = (
 
 
 def _is_macho_file(filename: str) -> bool:
-    """Return True if the file at `filename` begins with a Mach-O magic number."""
+    """Return True if file at `filename` begins with Mach-O magic number."""
     try:
         with open(filename, "rb") as f:
             header = f.read(4)
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -148,6 +148,30 @@ def _run(
         ) from exc
 
 
+# Mach-O magic numbers. See:
+# https://github.com/apple-oss-distributions/xnu/blob/5c2921b07a2480ab43ec66f5b9e41cb872bc554f/EXTERNAL_HEADERS/mach-o/loader.h#L65
+# https://github.com/apple-oss-distributions/cctools/blob/658da8c66b4e184458f9c810deca9f6428a773a5/include/mach-o/fat.h#L48
+MACHO_MAGIC = (
+    0xFEEDFACE.to_bytes(4, "little"),  # MH_MAGIC
+    0xFEEDFACE.to_bytes(4, "big"),  # MH_MAGIC
+    0xFEEDFACF.to_bytes(4, "little"),  # MH_MAGIC_64
+    0xFEEDFACF.to_bytes(4, "big"),  # MH_MAGIC_64
+    0xCAFEBABE.to_bytes(4, "big"),  # FAT_MAGIC (always big-endian)
+    0xCAFEBABF.to_bytes(4, "big"),  # FAT_MAGIC_64 (always big-endian)
+)
+
+
+def _is_macho_file(filename: str) -> bool:
+    try:
+        with open(filename, "rb") as f:
+            header = f.read(4)
+            return header in MACHO_MAGIC
+    except PermissionError:
+        return False
+    except FileNotFoundError:
+        return False
+
+
 def unique_by_index(sequence):
     """unique elements in `sequence` in the order in which they occur
 
@@ -514,6 +538,8 @@ def get_install_names(filename: str) -> Tuple[str, ...]:
     InstallNameError
         On any unexpected output from ``otool``.
     """
+    if not _is_macho_file(filename):
+        return ()
     otool = _run(["otool", "-L", filename], check=False)
     if not _line0_says_object(otool.stdout or otool.stderr, filename):
         return ()
@@ -576,6 +602,8 @@ def _get_install_ids(filename: str) -> Dict[str, str]:
     InstallNameError
         On any unexpected output from ``otool``.
     """
+    if not _is_macho_file(filename):
+        return {}
     otool = _run(["otool", "-D", filename], check=False)
     if not _line0_says_object(otool.stdout or otool.stderr, filename):
         return {}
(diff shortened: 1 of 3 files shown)


delocate/tools.py
```diff
@@ -148,6 +148,30 @@ def _run(
         ) from exc
 
 
+# Mach-O magic numbers. See:
+# https://github.com/apple-oss-distributions/xnu/blob/5c2921b07a2480ab43ec66f5b9e41cb872bc554f/EXTERNAL_HEADERS/mach-o/loader.h#L65
```
matthew-brett, line 152: Thanks for this - is there anything else you know of that gives Apple's imprimatur to these numbers for some forseeable future?    I see these are official Apple repositories, but it would be reassuring to see some official documentation too.

delocate/tools.py
```diff
@@ -148,6 +148,30 @@ def _run(
...
+# https://github.com/apple-oss-distributions/cctools/blob/658da8c66b4e184458f9c810deca9f6428a773a5/include/mach-o/fat.h#L48
+MACHO_MAGIC = (
+    0xFEEDFACE.to_bytes(4, "little"),  # MH_MAGIC
+    0xFEEDFACE.to_bytes(4, "big"),  # MH_MAGIC
+    0xFEEDFACF.to_bytes(4, "little"),  # MH_MAGIC_64
+    0xFEEDFACF.to_bytes(4, "big"),  # MH_MAGIC_64
+    0xCAFEBABE.to_bytes(4, "big"),  # FAT_MAGIC (always big-endian)
+    0xCAFEBABF.to_bytes(4, "big"),  # FAT_MAGIC_64 (always big-endian)
```
matthew-brett, line 160: I suppose FAT applies to both 32/64 bit Intel dual-arch, and 64 bit Intel/64 bit ARM dual arch?

delocate/tools.py
colesbury: I'm not able to find any detailed Apple documentation for the mach-o format other than the header files. There are a variety of non-Apple blog posts and that describe the format, although none seemed sufficiently thorough to be included here.

delocate/tools.py
colesbury: Yes. Apple also previously used it for the PowerPC/Intel fat binaries ("universal binaries"). According to wikipedia, NeXT also used the same format for dual m68k/IA-32 in the 1990s.

matthew-brett: Great - thanks.

delocate/tools.py
matthew-brett: OK - if that's the best we can do, it's not too bad.

delocate/tools.py
```diff
@@ -148,6 +148,30 @@ def _run(
...
+)
+
+
+def _is_macho_file(filename: str) -> bool:
```
matthew-brett, line 164: Would you mind running this over some actual `.so` files?   Maybe use the files I added in this branch?  https://github.com/matthew-brett/delocate/tree/test-files 

See the `delocate/tests/data` directory for the `np-*.dylib` files.

colesbury: I added a test in [640d0cb](https://github.com/matthew-brett/delocate/pull/172/commits/640d0cb12859a16a8c79e6c445d0afb37ac36c95) that only the mach-o files in test-data are counted as such. If you merge your test-files branch, they'll be automatically tested. I also ran the test manually with those files.

Of course, `_is_macho_file` just looks at the first 4 bytes so non-macho files may be reported as macho files. (I think java bytecode files also use the 0xcafebabe magic number.) That's fine though, it just means more calls to `otool`.

HexDecimal, line 164: Maybe add a short docstring to `_is_macho_file` to say what it actually does:
```suggestion
def _is_macho_file(filename: str) -> bool:
    """Return True if the file at `filename` begins with a Mach-O magic number."""
```

matthew-brett:
Pending the CI - this looks good to me.  As you point out, the check is just a screening test to reject implausible files - we still run otool on everthing that survives this screening test.  Our only worry would be if there was some binary out there where the screen would reject it - where the beginning bytes are not in the listed patterns.

@HexDecimal - any further thoughts?

delocate/tools.py
```diff
@@ -148,6 +148,31 @@ def _run(
...
+    """Return True if the file at `filename` begins with a Mach-O magic number."""
```
matthew-brett, line 165:
```suggestion
    """Return True if file at `filename` begins with Mach-O magic number."""
```

HexDecimal:
> any further thoughts?

With how `MACHO_MAGIC` is currently used it could be better to store it as a `frozenset`.  This will speed up the `header in MACHO_MAGIC` check by a small amount.

Since the files are opened for only the first 4 bytes I'd wonder if file buffering should be disabled.  I don't know the actual performance implications of doing this.

I can't do profiling so I don't know what's worth focusing on.  Performance might be fine as-is.

matthew-brett:
@HexDecimal - if you can't see a logical error here - let's merge as is - we may be able to improve the performance later, but it's already a huge speedup.

HexDecimal (approved):

delocate/tools.py
```diff
@@ -148,6 +148,31 @@ def _run(
         ) from exc
 
 
+# Mach-O magic numbers. See:
+# https://github.com/apple-oss-distributions/xnu/blob/5c2921b07a2480ab43ec66f5b9e41cb872bc554f/EXTERNAL_HEADERS/mach-o/loader.h#L65
+# https://github.com/apple-oss-distributions/cctools/blob/658da8c66b4e184458f9c810deca9f6428a773a5/include/mach-o/fat.h#L48
+MACHO_MAGIC = (
```
matthew-brett, line 154: Any interest in making this a Frozenset as per suggestion by @HexDecimal ?

colesbury: Done

matthew-brett:
OK - thanks for your patience - in it goes.

PR #177, opened 2023-01-27
MRG: Add Kyle to LICENSE

matthew-brett:


matthew-brett:
@HexDecimal - OK with this?

HexDecimal:
Seems unnecessary since the license is already permissive and specific code ownership can be tracked with the Git history.  I might suggest something like `Matthew Brett and the Delocate contributors` instead.

matthew-brett:
How about now?

HexDecimal (approved):

PR #182, opened 2023-05-09
Suppress filename output in unzip call.

### LINKED ISSUE #181: `delocate-wheel` fails on UnicodeDecodeError
### Description

See title. The command (`delocate-wheel`) errors out with `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd1 in position 396537: invalid continuation byte`, see log.

There have been no changes to this package recently, so it might be an environmental thing, but the Github virtual env changelog doesn't show an obvious change. Any ideas?

### Build log

https://github.com/arbor-sim/arbor/actions/runs/4904892311/jobs/8758317425#step:7:10823

### CI config

https://github.com/arbor-sim/arbor/actions/runs/4904892311/workflow

HexDecimal:
The `unzip` command on MacOS does not seem to handle Python's UTF-8 locale correctly. Running `unzip -q` should skip outputting these non-ASCII filenames in otherwise working runs.

In theory this change may also improve the performance of large archives.

Fixes #181

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -65,7 +65,7 @@ jobs:
         run: isort . --check --diff
 
   tests:
-    runs-on: macos-10.15
+    runs-on: macos-11
     steps:
       - uses: actions/checkout@v3
         with:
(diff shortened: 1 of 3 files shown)


HexDecimal:
@matthew-brett GitHub Actions has dropped support for `macos-10.15`, this will affect test coverage.

HexDecimal:
Merged since tests can't run with this PR and the fix has been verified.

PR #183, opened 2023-05-12
Refactor zip functions.

HexDecimal:
It was suggested [here](https://github.com/matthew-brett/delocate/issues/181#issuecomment-1544388928) to port the zip functions from auditwheel.
Changes were made to fit Delocate's test conditions and use-cases and to clean up these functions.

Calls to the `unzip` command were removed, Windows can now test these functions.

### DIFF
--- a/delocate/tools.py
+++ b/delocate/tools.py
@@ -5,12 +5,13 @@ import os
 import re
 import stat
 import subprocess
-import time
 import warnings
 import zipfile
+from datetime import datetime
+from os import PathLike
 from os.path import exists, isdir
 from os.path import join as pjoin
-from os.path import relpath
+from pathlib import Path
 from typing import (
     Any,
     Dict,
@@ -812,22 +813,42 @@ def add_rpath(filename: str, newpath: str, ad_hoc_sign: bool = True) -> None:
         replace_signature(filename, "-")
 
 
-def zip2dir(zip_fname: str, out_dir: str) -> None:
+def zip2dir(
+    zip_fname: str | PathLike[str], out_dir: str | PathLike[str]
+) -> None:
     """Extract `zip_fname` into output directory `out_dir`
 
     Parameters
     ----------
-    zip_fname : str
+    zip_fname : str or Path-like
         Filename of zip archive to write
-    out_dir : str
+    out_dir : str or Path-like
         Directory path containing files to go in the zip archive
     """
-    # Use unzip command rather than zipfile module to preserve permissions
+    # The zipfile module does not preserve permissions correctly
     # http://bugs.python.org/issue15795
-    _run(["unzip", "-q", "-o", "-d", out_dir, zip_fname], check=True)
+    with zipfile.ZipFile(zip_fname, "r") as zip:
+        for name in zip.namelist():
+            member = zip.getinfo(name)
+            extracted_path = zip.extract(member, out_dir)
+            attr = member.external_attr >> 16
+            if member.is_dir():
+                os.chmod(extracted_path, 0o755)
+            elif attr != 0:
+                mode = attr & 0o777  # Permission bits
+                os.chmod(extracted_path, mode)
+            # Restore timestamp
+            modified_time = datetime(*member.date_time).timestamp()
+            os.utime(extracted_path, (modified_time, modified_time))
 
 
-def dir2zip(in_dir, zip_fname):
+def dir2zip(
+    in_dir: str | PathLike[str],
+    zip_fname: str | PathLike[str],
+    *,
+    compression: int = zipfile.ZIP_DEFLATED,
+    compress_level: int = -1,
+) -> None:
     """Make a zip file `zip_fname` with contents of directory `in_dir`
 
     The recorded filenames are relative to `in_dir`, so doing a standard zip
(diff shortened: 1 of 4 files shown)


delocate/tools.py
```diff
@@ -836,33 +857,34 @@ def dir2zip(in_dir, zip_fname):
...
-            z.writestr(info, contents, zipfile.ZIP_DEFLATED)
-    z.close()
+    with zipfile.ZipFile(
+        zip_fname, "w", compression=compression, compresslevel=compress_level
+    ) as zip:
+        for root, dirs, files in os.walk(in_dir):
+            for dir in dirs:
+                dir_path = Path(root, dir)
+                out_dir_name = str(dir_path.relative_to(in_dir)) + "/"
+                zip_info = zipfile.ZipInfo.from_file(dir_path, out_dir_name)
```
matthew-brett, line 876: I assume this normalizes the path to Unix style?

delocate/tools.py
```diff
@@ -812,22 +813,42 @@ def add_rpath(filename: str, newpath: str, ad_hoc_sign: bool = True) -> None:
...
         Directory path containing files to go in the zip archive
     """
-    # Use unzip command rather than zipfile module to preserve permissions
+    # The zipfile module does not preserve permissions correctly
     # http://bugs.python.org/issue15795
-    _run(["unzip", "-q", "-o", "-d", out_dir, zip_fname], check=True)
-
-
-def dir2zip(in_dir, zip_fname):
+    with zipfile.ZipFile(zip_fname, "r") as zip:
```
matthew-brett, line 830: Is this code from auditwheel?

delocate/tools.py
```diff
@@ -812,22 +813,42 @@ def add_rpath(filename: str, newpath: str, ad_hoc_sign: bool = True) -> None:
...
+        for name in zip.namelist():
+            member = zip.getinfo(name)
+            extracted_path = zip.extract(member, out_dir)
+            attr = member.external_attr >> 16
```
matthew-brett, line 834: How does this logic relate to the results of `unzip`?

matthew-brett (commented):
Thanks - sorry - just some driveby comments.

delocate/tools.py
HexDecimal: I just now checked: Yes. `zipfile.ZipInfo.from_file` always calls [os.path.normpath](https://docs.python.org/3/library/os.path.html#os.path.normpath) on the path given.
> On Windows, it converts forward slashes to backward slashes.

delocate/tools.py
HexDecimal: `zip2dir` and `dir2zip` were taken from auditwheel, [specifically from here](https://github.com/pypa/auditwheel/blob/321292d620f14ddc7ad3247d71746679d6e51558/src/auditwheel/tools.py#L31-L91), and then modified.  Delocate has different requirements from permission bits and timestamps.

The docstrings and parameters are very similar which makes it look like it's been copied before, in one direction or the other or from the same sources.

delocate/tools.py
HexDecimal: `zip.extract` does a pretty good at preserving file attributes, [but it doesn't do a perfect job due to its known issues with file permissions](http://bugs.python.org/issue15795).

After the file is extracted the rest of the code is to fix the permissions and preserve the timestamp.

delocate/tools.py
HexDecimal: I accidentally read this backwards.  This would actually imply that it always converts to Windows separators on Windows.

HexDecimal: Once the archive path is passed to `ZipInfo.__init__` it converts the slashes:
```py
# This is used to ensure paths in generated ZIP files always use
# forward slashes as the directory separator, as required by the
# ZIP format specification.
if os.sep != "/" and os.sep in filename:
    filename = filename.replace(os.sep, "/")
```

delocate/tools.py
matthew-brett: Are the relationships between the `external_attr` and desired output permissions written up somewhere that we could cite?   It would be good to read the code and see why permissions are in the most significant 2 bytes of the int, why directories are always 755, why 0 means leave permissions as they are...

HexDecimal: > Are the relationships between the external_attr and desired output permissions written up somewhere that we could cite?

Probably not, since when I try to look it up I find that `external_attr`'s poor documentation is a known issue:
https://github.com/python/cpython/issues/62462

I've been assuming theses are internals to the zip format itself.

Also not sure why directories are always 755, going back to auditwheel and following the blame logs for that code doesn't give an explanation.  So I just take it at face value as making sure the directory has enough permissions to continue extracting files and maybe running them.

matthew-brett: Maybe add comment and that link to let people know that this is an inherited heuristic, and in the hope that better documentation surfaces at some point.

HexDecimal: You mean `external_attr`?  It seems it's platform dependent too, with different platforms setting different bits, which explains why the high bits were compared to zero since a non-Unix platform probably won't even set them.

The most information I found on it was here: https://unix.stackexchange.com/questions/14705/the-zip-formats-external-file-attribute

HexDecimal: I've changed to better variable names and added a link to the most useful resource I could find on external_attr.

Issue #125, opened 2021-09-21
Code cleanup suggestions.


HexDecimal:
Some ideas that could make things easier for me and future maintainers:

I'd be nice to setup auto formatters like Black and isort.  It's tedious to format all code manually.

[doc/devel/make_release.rst](https://github.com/matthew-brett/delocate/blob/dbb9367bcf2b8519199587b5336bcbd2d12e3c82/doc/devel/make_release.rst) should have been in a `CONTRIBUTING` file.  I missed it entirely when I made the `0.9.1` release.  A `CONTRIBUTING` file could also be used to explain how to add any test data.

In #93 I attempted to fully type-hint the package.  I never got that merged but it showed some areas that are causing issues such as `InWheelCtx.__enter__` and `back_tick`.  I think `InWheelCtx` needs to be refactored or removed (or finish #46 and fix things there,) and all calls to `back_tick` should be replaced with `subprocess.run` now that the code base is on Python 3.

Testing has some issues.  Everything from `pytest_tools.py` breaks the pytest debugger so they'll all eventually need to be replaced with plain asserts.  It'd be nice to replace `scriptrunner.py` with a more standard tool like [pytest-console-scripts](https://pypi.org/project/pytest-console-scripts/).

Issue and PR templates might be a good idea.  It sucks that most of the current bug reports don't mention which version of decloate they used and don't provide a wheel that can be checked.

matthew-brett:
All good suggestions.   Would you like me to review #93?   I suppose we should keep / deprecate `backticks` for a bit, just in case, but yes, all uses should be replaced with `subprocess.run`.

PR #185, opened 2023-05-12
Replace ScriptRunner with pytest-console-scripts fixtures.

HexDecimal:
This resolves one of the tasks in #125

I'm throwing out `scriptrunner.py`.  Any desired features can be implemented upstream.  I've added my own protocols to get around its missing typing.

Updated type-hints and asserts for all the scripts tests.  There was more I wanted to do but my changes were beginning to reach into other modules.


matthew-brett:
Looks good - did you check that the script tests fail if the scripts are broken in some way?

HexDecimal:
> Looks good - did you check that the script tests fail if the scripts are broken in some way?

This does what the old script runner was doing but it's maintained upstream and the new script runner is well tested.  I've even added my own features and tests (and of course full type-hinting) to the fixture upstream.

Some tests were exact so I'm not sure what you mean by breaking them.  Scripts which were expected to fail before still fail as expected as their returncode is checked.  There's some changes with the new outputs since this returns strings like subprocess.run instead of lists of strings which were common with the old runner and that will have changed the tests a little.

matthew-brett:
Sorry - I only meant - did you try breaking the code for the scripts and running the tests to make sure they fail in that circumstance?

HexDecimal:
> did you try breaking the code for the scripts and running the tests to make sure they fail in that circumstance?

I didn't do that. All of the Delocate script tests use MacOS linkage so I can't break them locally.  The tests were all broken when the new script runner wasn't returning lists of strings anymore, but that was mostly a process of making the type checker happy rather than waiting on the CI to give me test results.  Then after that the tests still pass and test coverage hasn't even changed despite me deleting `delocate/tests/scriptrunner.py`, so I see that as the new fixture being a perfect replacement for the bundled one.

I did notice that `run_command` checks the return code and raises on failure so I made sure that `check=True` was added to the new fixtures, but that was a new feature I just added upstream after starting this PR.  There are already scripts which cause a failure and raise exceptions and those were already tested but I went and cleaned up the code more.

If there's anything more specific you want to try then you're in a better position than I am to prod at it.

HexDecimal:
I've tested "breaking the code for the scripts" and the results are as expected: tests fail with verbose output of the failing command including the error code, stdout, and stderr.

matthew-brett:
Thanks for this ...

HexDecimal:
I assume you're too busy to handle these PR's? This PR was blocking me from working on more script related tasks. I'd like to merge 2 other PR's as well (the ones that were requested, which I've self-assigned.)

matthew-brett:
Yes, sorry - I'm mid-term and very busy at the moment.

Issue #139, opened 2022-01-01
Change default branch to `main`


matthew-brett:
Git has changed its default branch name to `main`, from `master`.   `main` seems like a reasonable name for the main development branch.   Other projects, such as Scipy, are also making this switch.   I propose we do this too.

The only possible issue is Versioneer - I see it hard-codes `master` in some code, but I'm not sure whether this will have any practical impact on us.

matthew-brett:
I'm tempted to go ahead with this on the Github interface, and then clean up any breakage.  Any objections?

HexDecimal:
I'm okay with it happening at anytime. I personally don't care about the related social drama, I just like the shorter name better.

PR #186, opened 2023-05-22
Replace versioneer with setuptools-scm, migrate to pyproject.toml.

HexDecimal:
#139 mentioned issues with Versioneer, so this PR completely removes it.  `pyproject.toml` allows including versioning tools into the build system so having a bundled addon is no longer necessary.

Now `_version.py` is generated during setup, but I've tweaked the script so that the package can still run without it.

A lot of setup.py metadata has been moved into pyproject.toml as well. With a little more effort I think the entire thing can be moved to pyproject.toml but I stuck with the more important bits for now.

With this change you no longer use `setup.py` to build or develop.  You use `pip install -e .` to install in development mode and `python -m build` to make distributions. I've automated this stuff some time ago and have now edited the release guide to reflect that.

If this PR does too much then I can split it up.


delocate/__init__.py
```diff
@@ -1,11 +1,12 @@
 # Init for delocate package
 
-from . import _version
 from .delocating import delocate_path, delocate_wheel, patch_wheel
 from .libsana import tree_libs, wheel_libs
 
-__version__ = _version.get_versions()["version"]
-del _version
+try:
```
matthew-brett, line 6: Sorry for my ignorance here, but under what circumstances would this be triggered?

HexDecimal: Not under normal circumstances, but if the module was imported directly from the repo without being installed in development mode then `_version.py` would be missing.  Versioneer seemed to have a fallback for the case when a version was asked for with the package being setup and this was me accounting for that with the new setup.

It should also be possible to cover this with a test which installs the dependencies but not the package itself. I'd also be okay with removing this try/except clause and just telling everyone to be sure to use `pip install -e .`

matthew-brett: How about raising a warning in the block, possibly a deprecation warning?

.github/workflows/pypi-deploy.yml
```diff
@@ -9,17 +9,17 @@ jobs:
   deploy:
     runs-on: ubuntu-latest
     steps:
-      # checkout@v1 is required since versioneer uses git describe.
+      # checkout@v1 is used to that tags are more relaible.
```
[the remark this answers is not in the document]
HexDecimal: The checkout action has a history of messing up references to annotated tags which could affect scripts deriving versions from those tags but if the build system is just using any tag then it will probably work fine.  https://github.com/actions/checkout/issues/290

I think I can update the version here without issues, but you don't really know until you push a commit and it breaks.

delocate/__init__.py
HexDecimal: A RuntimeWarning might be appropriate, but also simply crashing when `_version.py` is missing is also an option.  I'll go with whatever you want.

I think running the module without doing a development install is unrealistic and is going to crash on one of the imports of dependencies before it reaches this part.

I've had one of my own projects which required a trick like this but it was exceptional: It needed to import the package for doc generation but the package was too heavy to install on the doc generating servers.

matthew-brett: My preference would be to allow it for now, with a deprecation warning that we will break in the future - just because somebody may be using that mis-feature at the moment.

.github/workflows/pypi-deploy.yml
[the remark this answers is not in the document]
matthew-brett: OK - I added the URL to the suggestion.

delocate/__init__.py
```diff
@@ -1,12 +1,20 @@
...
+import warnings
+
 from .delocating import delocate_path, delocate_wheel, patch_wheel
 from .libsana import tree_libs, wheel_libs
 
 try:
     from ._version import __version__
 except ImportError:
     __version__ = "0.0.0"
+    warnings.warn(
```
matthew-brett, line 12: But - what do you think about planning to raise an error after a period of deprecation?

delocate/__init__.py
```diff
@@ -1,12 +1,20 @@
...
+        "Delocate was not installed and is missing version metadata."
+        "\nMake sure this package is installed in development mode with this command:"
```
matthew-brett, line 14: PEP line length warning.

matthew-brett (commented):
Consider deprecation and planned error after deprecation?  Your call.  PEP complaint, otherwise all good.

delocate/__init__.py
matthew-brett: Your party, your rules!   Just go for whatever you prefer.

matthew-brett (approved):
Feel free to merge when you're happy ...

PR #187, opened 2023-08-25
Support for SOURCE_DATE_EPOCH (reproducible builds)

dalcinl:
https://reproducible-builds.org/docs/source-date-epoch/

### DIFF
--- a/delocate/tests/env_tools.py
+++ b/delocate/tests/env_tools.py
@@ -24,3 +24,20 @@ def TempDirWithoutEnvVars(*env_vars):
             else:
                 if var in os.environ:
                     del os.environ[var]
+
+
+@contextmanager
+def _scope_env(**env):
+    """Add `env` to the environment and remove them after testing
+    is complete.
+    """
+    env_save = {key: os.environ.get(key) for key in env}
+    try:
+        os.environ.update(env)
+        yield
+    finally:
+        for key, value in env_save.items():
+            if value is None:
+                del os.environ[key]
+            else:
+                os.environ[key] = value
--- a/delocate/tests/env_tools.py
+++ b/delocate/tests/env_tools.py
@@ -1,5 +1,6 @@
 import os
 from contextlib import contextmanager
+from typing import Iterator
 
 from ..tmpdirs import InTemporaryDirectory
 
@@ -27,7 +28,7 @@ def TempDirWithoutEnvVars(*env_vars):
 
 
 @contextmanager
-def _scope_env(**env):
+def _scope_env(**env: str) -> Iterator[None]:
     """Add `env` to the environment and remove them after testing
     is complete.
     """
--- a/delocate/tests/test_wheelies.py
+++ b/delocate/tests/test_wheelies.py
@@ -443,7 +443,7 @@ def test_fix_namespace() -> None:
         assert delocate_wheel(NAMESPACE_WHEEL, "out.whl") == stray_libs
 
 
-def test_source_date_epoch():
+def test_source_date_epoch() -> None:
     with InTemporaryDirectory():
         zip2dir(PURE_WHEEL, "package")
         for date_time, sde in (
--- a/delocate/tests/test_wheelies.py
+++ b/delocate/tests/test_wheelies.py
@@ -5,6 +5,7 @@ import shutil
 import stat
 import subprocess
 import sys
+import zipfile
 from glob import glob
 from os.path import abspath, basename, exists, isdir
 from os.path import join as pjoin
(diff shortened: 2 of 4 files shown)


dalcinl:
I'm not familiar enough with the codebase to add tests. Any tips would be most appreciated.

dalcinl:
coverage will not pass unless I add tests, but I would need directions on the best course of action.
Otherwise, I could rewrite my code in such a way that coverage would pass, but that would be cheating.

HexDecimal:
I'm not sure. I guess a test could be added for calling `delocate_wheel` on a copy of `PURE_WHEEL` with and without this environment variable set, see `test_wheelies.py`. Maybe copy the `test_fix_pure_python` test.

delocate/tools.py
```diff
@@ -844,12 +845,27 @@ def zip2dir(
             os.utime(extracted_path, (modified_time, modified_time))
 
 
+ZIP_TIMESTAMP_MIN = 315532800  # 1980-01-01 00:00:00 UTC
+DateTuple = Tuple[int, int, int, int, int, int]
+
+
+def get_zip_datetime(
```
HexDecimal, line 852: I'd like this to be a private function (add a single leading underscore.)  You should also add a short docstring to clarify when it returns what value, and maybe link to the standard.

delocate/tests/test_wheelies.py
```diff
@@ -439,3 +440,27 @@ def test_fix_namespace() -> None:
...
+                dir2zip("package", "package.zip")
+                with zipfile.ZipFile("package.zip", "r") as zip:
+                    for name in zip.namelist():
+                        member = zip.getinfo(name)
+                        assert_equal(member.date_time, date_time)
+        finally:
+            if sde_orig is None:
+                del os.environ["SOURCE_DATE_EPOCH"]
+            else:
+                os.environ["SOURCE_DATE_EPOCH"] = sde_orig
```
HexDecimal, line 466: The environment variable handing is all over the place. Would you mind rewriting this to use [contextlib.contextmanager](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)?

HexDecimal:
The last thing to do would be to add the changes to `dir2zip` and the support for `SOURCE_DATE_EPOCH` to the changelog file.

dalcinl:
I should mention that native macOS tools (e.g. `ld64`) do not really support `SOURCE_DATE_EPOCH`. They do support `ZERO_AR_DATE`, though. The behavior is different, they just zero timestamps (or set them -1, don't remember exactly).
Perhaps it would be nice to assume that `ZERO_AR_DATE` means the user has an intention to get reproducible binaries including the outer wheel file. With the code I put in place, the implementation would be a couple extra lines. I'll leave this nit to the consideration of the maintainers.

dalcinl:
> The last thing to do would be to add the changes to `dir2zip`

Do you mean to add to the changelog that `dir2zip` now has an extra optional keyword-only argument? Or something else? IMHO, it is a quite trivial and inconsequential addition not worth of a changelog entry. Last time `dir2zip` was modified, new kw-only args were added, and yet no entry in the changelog for such change.

HexDecimal:
> Do you mean to add to the changelog that dir2zip now has an extra optional keyword-only argument?

I did mean that, but it's vague how much changes in public functions should be compared the actual features and CLI arguments. Personally, I'd rather consider all but the most public facing tools to be private, but you can never be sure unless you've added that leading underscore to everything ahead of time. I'm not sure there's consensus on this at the moment.

delocate/tests/env_tools.py
```diff
@@ -24,3 +24,17 @@ def TempDirWithoutEnvVars(*env_vars):
             else:
                 if var in os.environ:
                     del os.environ[var]
+
+
+@contextmanager
+def Environment(**env):
```
HexDecimal, line 30: Don't name this like a class, add annotations, and a docstring, maybe also make it private.
```suggestion
def _scope_env(**env: str) -> Iterator[None]:
    """Add the provided environment variables for the duration of this context."""
```

dalcinl:
> but it's vague how much changes in public functions

FWIW, a new kw-only argument (in a function without `**kwargs`) has absolutely zero backward compatibility implications.

> Personally, I'd rather consider all but the most public facing tools to be private

That's what I assumed for this particular project, and why I did not care in my original sumision about making private the new utility function I added. This is a command line tool. A note in README should be enough to make the point clear.

delocate/tests/env_tools.py
dalcinl: > Don't name this like a class,

Done. I named the new context manager like a class precisely because of the other context manager function in the same file.

dalcinl: I missed the rename suggestion. I'll push a fix.

HexDecimal (approved):

PR #197, opened 2023-11-29
Expand globs in simple script arguments

### LINKED ISSUE #71: [BUG] When delocate is called with a non-expanded wildcard path it produces wrong result

HexDecimal:
Expands globs in simple cases using the following rules:

* If any path is a existing folder or file then it's used as-is, even if it can be used as a glob pattern. This is supposed to ensure that there's no surprises or regressions.
* Otherwise the glob is expanded with Python's `glob` module. If this results in no files then it raises `FileNotFoundError`.

Added tests for this.

Some commands are too complex for this workaround. This only applies to commands which takes one list of arguments and not commands with more complex positional parameters such as `delocate-fuse` and `delocate-patch`.

I'd rather not merge this PR until it's reviewed.

Fixes #71

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,7 +4,7 @@
     ],
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": {
-        "source.fixAll": true,
+        "source.fixAll": "always"
     },
     "[python]": {
         "editor.defaultFormatter": "charliermarsh.ruff"
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,12 @@ rules on making a good Changelog.
 
 ## [Unreleased]
 
+### Changed
+
+- Handle glob paths in more cases for `delocate-wheel`, `delocate-path`,
+  `delocate-listdeps`, and `delocate-addplat`.
+  [#71](https://github.com/matthew-brett/delocate/issues/71)
+
 ### Fixed
 
 - No longer picks a random directory when multiple directories end with the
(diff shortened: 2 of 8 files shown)


HexDecimal:
After some more thought and better tests I've decided that this shouldn't cause issues when merged. Preferring literal names to glob expansion should avoid theoretical regressions, and this PR is very easy to revert if any unexpected edge cases do appear.

PR #198, opened 2024-02-04
Validate and update macOS platform tag in wheel name

### LINKED ISSUE #56: Should MACOSX_DEPLOYMENT_TARGET/LC_VERSION_MIN_MACOSX be verified?
When building wheels for macOS target version 10.9 (for example), then delocate will happily pull in dependencies that have a higher target version. In most cases this won't cause issues, but I think there should at least be a warning.

From https://stackoverflow.com/a/17462763, this is how to show the target version from a lib:
```sh
$ otool -l foo.dylib | grep -A 3 LC_VERSION_MIN_MACOSX
      cmd LC_VERSION_MIN_MACOSX
  cmdsize 16
  version 10.8
      sdk 10.8
```

This will typically happen when using homebrew to get dependencies. The downloaded bottles will have the same target version as the host OS. Unfortunately, brew can't be forced to build packages from source with a custom target version: https://discourse.brew.sh/t/it-is-possible-to-build-packages-that-are-compatible-with-older-macos-versions/4421. The only correct solution then would be either to build wheels on an old macOS version that becomes the target version, or to not use homebrew and build dependencies manually while setting `MACOSX_DEPLOYMENT_TARGET` accordingly.

Czaki:
This PR introduces scanning all binaries in the wheel and checking if the platform tag in the wheel name is correct in the context of embedded binaries. 

By default, it updates the wheel name. If `--require-target-macos-version` is provided or `MACOSX_DEPLOYMENT_TARGET` is provided, then delocate fails if the wheel does not meet versions provided this way.


delocate/delocating.py
```diff
@@ -704,6 +830,16 @@ def delocate_wheel(
...
+        if check_wheel_name or fix_wheel_name:
+            wheel_name = os.path.basename(in_wheel)
+            new_name = _update_wheel_name(wheel_name, wheel_dir)
+            if check_wheel_name and new_name != wheel_name:
+                raise DelocationError(
+                    "Wheel name does not satisfy minimal package requirements"
+                )
+            if new_name != wheel_name:
+                in_place = False  # maybe something better
+                out_wheel = os.path.join(os.path.basename(out_wheel), new_name)
```
HexDecimal, line 842: Any chance this section can be moved into its own private function?

HexDecimal (changes requested):
I'm in a good position to review this, and merge it once it's finished.

The following is required:

* New CLI flags mentioned in changelog
* Test passing and failing wheel names for `--verify-name`
* Test changed and unchanged names for `--fix-name`

`test_scripts.py` would be the appropriate place to add these tests.

delocate/delocating.py
```diff
@@ -578,6 +591,113 @@ def _make_install_name_ids_unique(
...
+    # get the architecture and minimum macOS version from the libraries
+    # in the wheel
+    version_set = set()
+    for lib in glob.glob(
+        pjoin(wheel_dir, "**", "*.dylib"), recursive=True, include_hidden=True
+    ):
+        version_set.update(_get_macos_min_version(lib))
+    for lib in glob.glob(
+        pjoin(wheel_dir, "**", "*.so"), recursive=True, include_hidden=True
+    ):
```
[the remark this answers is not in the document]
Czaki: Changed to use Path.glob instead

delocate/delocating.py
```diff
@@ -704,6 +830,16 @@ def delocate_wheel(
             install_id_prefix=DLC_PREFIX + relpath(lib_sdir, wheel_dir),
         )
         rewrite_record(wheel_dir)
+        if check_wheel_name or fix_wheel_name:
+            wheel_name = os.path.basename(in_wheel)
+            new_name = _update_wheel_name(wheel_name, wheel_dir)
+            if check_wheel_name and new_name != wheel_name:
+                raise DelocationError(
+                    "Wheel name does not satisfy minimal package requirements"
+                )
```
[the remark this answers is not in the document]
Czaki: You mean to provide a list of files that have higher system required or only provide new and old tag? 

If first, should I return new name and list of problematic files or perform scan again to discover (it will require more code, but the function to determine wheel name will be more readable)?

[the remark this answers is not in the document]
HexDecimal: You're right, that could make it more complicated.

I was mostly worrying about people posting bug reports with this error message being the only context given. If there was a simple response to that kind of post then I'd like it to be included in the error message itself.

I was only thinking of the 2nd option since this code should already have that much information, but your 1st option would be much more helpful for anyone encountering this error. I'll accept either.

HexDecimal:
Minor issue with code coverage, simply re-running the tests solved it. Or were you asking for advice on writing tests?

Czaki:
> Or were you asking for advice on writing tests?

No. I do not want to start writings test before clarifying the interface.

Czaki:
One additional question. If both checks should be optional or some should be default behavior with option to disable?

HexDecimal:
> One additional question. If both checks should be optional or some should be default behavior with option to disable?

If you made this not optional that would simplify the code. There would be no need to verify the versions anymore if this always sets the version correctly. No need for the additional CLI flags either. I can't imagine use cases where anyone would want to keep the wrong version intentionally, but feel free to correct me.

I looked at issue #56 but it appears you're the one who made that issue. I don't have enough MacOS experience to add much additional insight. Maybe reduce the checks suggested here to only `--require-target-macos-version <max_version>` which errors if an upper bound is exceeded. I'll assume this PR closes #56.

To clarify what I'm suggesting:

* Always rename wheels to the current minimum supported version regardless of flags given.
* Verify minimum version compared to `--require-target-macos-version <max_version>`  when provided. If verification fails then output which libraries are the cause and which versions they wanted.
* Remove `--verify-name` and `--fix-name`. Fixing is always implied and verification needs a target version to compare to which can't be determined automatically.

If you want an error message format, how about this:
```
Library dependencies do not satisfy target MacOS version {require_target_macos_version}:
{lib_path} has a minimum target of {lib_macos_version}
...  # repeat for each failed library
```

delocate/delocating.py
```diff
@@ -578,6 +591,191 @@ def _make_install_name_ids_unique(
...
+    result = set()
+    for v, lib in version_lib_dict.items():
+        if v > version:
+            result.update(lib)
+    return result
+
+
+def _calculate_minimum_wheel_name(
+    wheel_name: str, wheel_dir: Path
+) -> Tuple[str, set[Path]]:
```
HexDecimal, line 685: Replace the Python 2 `from __future__ import division, print_function` with `from __future__ import annotations`. This will let you use PEP 585 generics in type-hints.

Czaki:
I have applied requested changes. I have added the option to activate `require-target-macos-version` by setting `MACOSX_DEPLOYMENT_TARGET` environment variable.

HexDecimal:
> If wheel name does not fit the proper pattern, should run fail (and I need to fix all tests) or it should skip name validation then?

Wheel names must be valid after renaming, there are PEP's for what is a valid name. Wheel inputs should be valid enough for tools to work. I'll look into the test errors and the wheel spec to figure out why tests are currently failing.

> Should I propose changes in README about new error?

The most important place is the `CHANGELOG`. It should note the new CLI flag and that wheels might be tagged with a different MacOS version than before. It should also note that new ENV variables will be read.

You can also explain these in the `README`, but the `CHANGELOG` is more important since there needs to be an easy to read record of when these changes have happened.

HexDecimal:
It seems that current tests assume wheels can be given arbitrary names when used with Delocate, and your new code assumes wheels have a full valid name to work with to pull tags from the filename.  I think your code renaming files has broken the assumption made by Delocate that wheels will always have the same name before and after delocation and that the actual name doesn't matter.

After some consideration I think Delocate is wrong to have this assumption, and I think it's safe to force wheels to have valid input names. This will require rewriting some tests. This might break the tool for users abusing wheel names, but regular users will be unaffected. Stricter inputs will need to be mentioned in the `CHANGELOG`.

Wheels have tag information included in the wheel metadata, this can be read to get the tags of a wheel with an invalid name. I assume it must be updated when the tags of a wheel are changed as well. The metadata location is based on a valid wheel name itself so it isn't an alternative to a valid name.

More info: https://packaging.python.org/en/latest/specifications/binary-distribution-format/

I think Delocate might be breaking several packaging rules, but I'd have to take a longer look to know for sure how bad it is.

> If wheel name does not fit the proper pattern, should run fail (and I need to fix all tests) or it should skip name validation then?

With all this in mind, the current tests are broken and need to be fixed by giving temporary wheels valid names (or by putting the copied wheels into their own temporary directories so they that don't conflict with each other.)

Czaki:
Ok. I have added tests and changed behavior to requested. 

I do not have an idea how properly test `test_check_plat_archs` (failing currently). 

As the parser is defined globaly in `delocate.cmd.delocate_wheel` it is not easy to run `test_delocate_wheel_verify_name_universal2_verify_crash` with `MACOSX_DEPLOYMENT_TARGET` environment variable. It could be fixed by moving the parser definition to a function or reload module using `importlib.reload`.

HexDecimal:
> test_delocate_wheel_verify_name_universal2_verify_crash with MACOSX_DEPLOYMENT_TARGET environment variable. It could be fixed by moving the parser definition to a function or reload module using importlib.reload.

`pytest-console-scripts` has a setting to handle this case:
```py
@pytest.mark.script_launch_mode('subprocess')
def test_delocate_wheel_verify_name_universal2_verify_crash(...
```
This will cause `script_runner.run` to run in a new process, causing the import-time code to run again. Monkey patches will no longer work, so make a copy with `os.environ.copy()`, then set the tested variable, then pass the modified environment to `script_runner.run(env=)`

HexDecimal:
> I do not have an idea how properly test test_check_plat_archs

I suspect there's nothing wrong with `test_check_plat_archs`. It seems `_calculate_minimum_wheel_name` is having trouble with `fakepkg1-1.0-cp37-abi3-macosx_10_9_universal2.whl`. It looks like `_get_macos_min_version` doesn't return `arm64`, then the function crashes later when it checks for `arm64` on the `universal2` wheel.

It's possible that `_get_macos_min_version` and `_get_archs_and_version_from_wheel_name` will need their own tests to ensure they work as documented.

HexDecimal:
> Yes. Because in this test, the arm64 binaries were removed from the wheel. The question is how to fix this. I need to add some additional binaries to the default wheel? As I expect that delocate should fail if there is no binary for given architecture.

Thanks for clarifying, I misunderstood what I was looking for.

This test appears to be for the `require_archs` parameter of `delocate_wheel` (also the `--require-archs` flag). With the failing code running `require_archs=()`. The relevant documentation for this argument:

> An empty sequence results in checks that depended libraries have the same archs as depending libraries.

So this ignores the wheels tags and verifies the bundled libraries as-is. The old code handles wheels with invalid architecture combinations without fail, but your code is more strict and crashes when a wheel doesn't contain the correct architectures based on its tag. I prefer the strictness, but it would've been nice if the error was better than KeyError.

This test has the terribly documented `_fix_break` and `_fix_break_fix` sub-functions. These can be updated to return renamed wheels with the correct arch tags so that they won't be invalid anymore. This is a small and easy change compared to any other idea I can think of.

Czaki:
Looks ready for review.

Changelog.md
```diff
@@ -10,6 +10,18 @@ rules on making a good Changelog.
...
+- Starting from this release, the 'delocate' tool will use the content
+  of the wheel file to determine the platform tag, which will be included in the wheel's name.
+  This change is intended to prevent confusion related to the platform tag.
+  The 'delocate-wheel' tool is now required to adjust the wheel's name appropriately.
+  If the `--require-target-macos-version` option is used or if the `MACOSX_DEPLOYMENT_TARGET`
+  environment variable is set, 'delocate-wheel' will enforce that the wheel
+  is built for the specified macOS version.
+  If the wheel is not compatible with the specified macOS version,
+  `delocate-wheel` will terminate with an error.
+  This change is part of pull request #198.
```
HexDecimal, line 23: One PR can add multiple items. The changelog is not the place to justify changes, so items should explain WHAT/HOW and not WHY. "Starting from this release" is implied by the position of items.

```suggestion
### Added

- Use `--require-target-macos-version` or the `MACOSX_DEPLOYMENT_TARGET`
  environment variable to ensure that wheels are
  compatible with the specified macOS version. #198

### Changed

- Delocate now uses the binaries of the wheel file to determine a more accurate
  platform tag, this will rename wheels accordingly. #198
- `delocate-wheel` is now more strict with platform tags and will no longer allow
  a wheel to be incompatible with its own tags. #198
```

HexDecimal (changes requested):
I assume the tags in `{distribution}-{version}.dist-info/WHEEL` were not updated yet. It is probably incorrect to leave them unchanged.

I think this currently leaves the original file alone if the name is changed. This will mess up the distribution and deployment of most projects. If this renames a wheel then the original must be deleted (if it's in the same directory).

Czaki:
> I assume the tags in `{distribution}-{version}.dist-info/WHEEL` were not updated yet. It is probably incorrect to leave them unchanged.

I miss this. Earlier. I have fixed it now. 

> I think this currently leaves the original file alone if the name is changed. This will mess up the distribution and deployment of most projects. If this renames a wheel then the original must be deleted (if it's in the same directory).

You were right. It is fixed now.

delocate/delocating.py
```diff
@@ -578,6 +591,228 @@ def _make_install_name_ids_unique(
...
+            for lib_path, lib_macos_version in problematic_files
+        )
+        raise DelocationError(
+            "Library dependencies do not satisfy target MacOS"
+            f" version {require_target_macos_version}:\n"
+            f"{problematic_files_str}"
+        )
+    if new_name != wheel_name:
+        wheel_path = os.path.join(os.path.dirname(wheel_path), new_name)
+    return wheel_path
```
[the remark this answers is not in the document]
Czaki: Fixed and added docstring

delocate/delocating.py
```diff
@@ -578,6 +592,274 @@ def _make_install_name_ids_unique(
...
+        A set of libraries that require a more modern macOS version than the
+        provided one.
+    """
+    if version is None:
+        return set()
+    result: set[tuple[Path, Version]] = set()
+    for v, lib in version_lib_dict.items():
+        if v > version:
+            result.update((x, v) for x in lib)
+    return result
```
HexDecimal, line 684: Names should be more verbose. `lib` was actually multiple libraries which was not implied by its name.
```suggestion
def _get_problematic_libs(
    required_version: Optional[Version], version_lib_dict: Dict[Version, List[Path]]
) -> set[tuple[Path, Version]]:
    """
    Filter libraries that require more modern macOS
    version than the provided one.

    Parameters
    ----------
    required_version : Version or None
        The expected minimum macOS version.
        If None, return an empty set.
    version_lib_dict : Dict[Version, List[Path]]
        A dictionary containing mapping from minimum macOS version to libraries
        that require that version.

    Returns
    -------
    set[tuple[Path, Version]]
        A set of libraries that require a more modern macOS version than the
        provided one.
    """
    if required_version is None:
        return set()
    bad_libraries: set[tuple[Path, Version]] = set()
    for library_version, libraries in version_lib_dict.items():
        if library_version > required_version:
            bad_libraries.update((path, library_version) for path in libraries)
    return bad_libraries
```

delocate/delocating.py
```diff
@@ -578,6 +592,274 @@ def _make_install_name_ids_unique(
...
+    ------
+    str
+        The CPU type.
+    Version
+        The minimum macOS version.
+    """
+    if dylib_path.is_dir() or not _is_macho_file(dylib_path):
+        return
+    m = MachO(dylib_path)
+    for header in m.headers:
```
HexDecimal, line 613:
```suggestion
    for header in MachO(dylib_path).headers:
```

delocate/delocating.py
```diff
@@ -578,6 +592,274 @@ def _make_install_name_ids_unique(
...
+        The path to the dylib file.
+
+    Yields
+    ------
+    str
+        The CPU type.
+    Version
+        The minimum macOS version.
+    """
+    if dylib_path.is_dir() or not _is_macho_file(dylib_path):
```
HexDecimal, line 610: Does `_is_macho_file` break if the path is a directory? If that's the case then the function should be updated.
```suggestion
    if not _is_macho_file(dylib_path):
```

HexDecimal (changes requested):

delocate/tests/test_scripts.py
```diff
@@ -643,18 +653,186 @@ def test_glob(
...
+
+    result = script_runner.run(
+        [
+            delocate_wheel_path,
+            whl_10_9,
+        ],
+        check=False,
+        cwd=tmp_path,
+        env={"MACOSX_DEPLOYMENT_TARGET": "10.9"},
+    )
```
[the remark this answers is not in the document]
Czaki: I do not understand what caused this problem. I jut hard hit into this.

delocate/cmd/delocate_wheel.py
```diff
@@ -61,6 +63,14 @@
...
     " 'x86_64,arm64')",
 )
+parser.add_argument(
+    "--require-target-macos-version",
+    type=Version,
+    help="Verify if platform tag in wheel name is proper",
+    default=Version(os.environ["MACOSX_DEPLOYMENT_TARGET"])
+    if os.environ.get("MACOSX_DEPLOYMENT_TARGET")
+    else None,
+)
```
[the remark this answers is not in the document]
Czaki: I am not convinced that overwriting `args.require_target_macos_version` is a good idea, so I introduced an additional variable.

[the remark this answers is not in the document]
HexDecimal: That's fine, but `args` is an `argparse.Namespace`. It only holds the parsed attributes, so it's flexible what you're allowed to do with it.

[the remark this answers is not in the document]
Czaki: For me, `args` should contain only the output of parsing user input. It is a personal feeling, but I could change it to an overwriting parameter, as it saves one line.

delocate/tests/test_scripts.py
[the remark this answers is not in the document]
HexDecimal: I don't like this workaround since it implies an incomplete installation of Delocate's scripts for testing, or an unexpected Python environment. I won't accept this workaround without a proper explanation of why the normal code doesn't work.

If you can't explain it then read `MACOSX_DEPLOYMENT_TARGET` at run-time instead of import-time so that this can run like any other test and then the workaround will no longer be needed. Otherwise, at least show me the error or traceback.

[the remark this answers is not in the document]
HexDecimal: My mistake, I've been reading outdated commits and missed that these issues are already resolved.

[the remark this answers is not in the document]
Czaki: Error is              
```
FileNotFoundError: [Errno 2] No such file or directory: 'delocate-wheel'
```

It looks like PATH is not properly propagated.

delocate/tests/test_scripts.py
```diff
@@ -577,23 +582,29 @@ def test_add_platforms(script_runner: ScriptRunner) -> None:
...
-        )
-        assert "libextfunc.dylib excluded" in result.stderr
-        with InWheel("test.whl"):
-            assert not Path("plat_pkg/fakepkg1/.dylibs").exists()
-        # We exclude a library that does not exist so we should behave normally
-        script_runner.run(
-            ["delocate-wheel", "-e", "doesnotexist", "test2.whl"], check=True
-        )
-        _check_wheel("test2.whl", ".dylibs")
+def test_fix_wheel_with_excluded_dylibs(script_runner: ScriptRunner, tmp_path):
```
HexDecimal, line 585:
```suggestion
def test_fix_wheel_with_excluded_dylibs(script_runner: ScriptRunner, tmp_path: Path) -> None:
```

HexDecimal (changes requested):
No major issues as far as I can tell. There's one last thing I'd like to change, after that I think this PR is finished.

HexDecimal (approved):

delocate/delocating.py
```diff
@@ -578,6 +592,275 @@ def _make_install_name_ids_unique(
...
+    Returns
+    -------
+    set[tuple[Path, Version]]
+        A set of libraries that require a more modern macOS version than the
+        provided one.
+    """
+    if required_version is None:
+        return set()
+    if arch == "arm64" and required_version < Version("11.0"):
+        required_version = Version("11.0")
```
Czaki, line 684: When checking the code again, I realized that without adding this check, the error output may be a little too verbose. 

Case. 
Someone compiled `unniversal2` wheel where all `x86_64` deps have targeted `10.9` and all except one `arm64` deps have targeted 11.0. The problematic one had target 12.0. Without this fix, all `arm64` deps will be listed as problematic. With these lines added (and modified call), only single problematic one. 

Should I add tests?

HexDecimal: I have trouble understanding this. I do not know or remember enough about the architecture versioning history. I'll trust your knowledge on this is correct.

Adding a test might help with explaining this to me. This is also handling a edge case which should have its own test.

Czaki: Short summary. 
The first macOS that supports arm64 processors is macOS 11.0. So no arm64 binary will be compatible with macOS from line 10.x 

Also, compilers will just simple bump the deployment target to 11.0 if the requested value is lower. 

Some time ago, wheel maintainers decided that universal2 wheels could have a platform target below 11.0 if all arm64 binaries are 11.0 compatibles. 

I have added a comment with an explanation of the function, could you check?

HexDecimal: That sounds right. It was a good idea to add this extra check. Everything so far looks good.

HexDecimal (approved):

delocate/delocating.py
```diff
@@ -578,6 +592,281 @@ def _make_install_name_ids_unique(
         validate_signature(lib)
 
 
+def _get_macos_min_version(dylib_path: Path) -> Iterator[tuple[str, Version]]:
+    """Get the minimum macOS version from a dylib file.
```
mara004, line 596: I was wondering if this could be turned into a public API that one could import from delocate?
It might be useful in our setup to autodetect the min verison for wheeltags (currently we are hardcoding it, not ideal). I would prefer to set the right tag directly instead of post-processing/replacing a wheel.

HexDecimal: I could review a PR to make it public, though the function mainly just grabs the version via `macholib`. There's not much to it to not simply copy it.

mara004: I see, thank you for the feedback. Admittedly I hadn't looked too closely, and had thought `MachO` might be something internal of delocate, but seeing as it is from macholib I agree it'll be best to use that directly.

Issue #29, opened 2017-12-13
Add support for Windows wheels


ghost:
Currently the process of building windows wheels is filled with too many hacks. I propose that delocate support windows in the following way:

1. The wheel is unpacked
2. For each PYD/DLL file in the wheel
   a) identify dependencies
   b) recursively add all dependencies to a list
3. Remove all DLL names that are already in the wheel
4. The remaining list constitutes "missing dependencies." These should be copied into a subfolder
5. For all python packages `AddDllDirectory` should be prepended to `__init__.py` just after the `__future__` imports.

xoviat:
Note that `PyInstaller` [already has][1] this functionality:

```python
    try:
        for a in args.filenames:
            for fn in glob.glob(a):
                imports = PyInstaller.depend.bindepend.getImports(fn)
                if is_win:
                    assemblies = PyInstaller.depend.bindepend.getAssemblies(fn)
                    imports.update([a.getid() for a in assemblies])
                print(fn, imports)
    except KeyboardInterrupt:
        raise SystemExit("Aborted by user request.")
```

We can't copy this functionality because it's GNU, but we could consider adding a dependency on Windows.

[1]: https://github.com/pyinstaller/pyinstaller/blob/79c881a2dd2c37965f46c337c59d2ce475e79274/PyInstaller/utils/cliutils/bindepend.py#L41-L50

anthrotype:
I was also wondering if there’s an equivalent of delocate or auditwheel for Windows, so stumbled on this. Did anything happen since then? (Unfortunately I’m not a Windows user so I can’t be of help.)

adang1345:
See https://github.com/adang1345/delvewheel, which is an equivalent for Windows that I wrote.

jvolkman:
@HexDecimal this issue is referring to delocating windows wheels, not delocating macos wheels on windows. As mentioned above, `delvewheel` exists now for the former so maybe you can close this issue.

HexDecimal:
You're right. I guess I was always confused by the title. I'll close this since Delocate is setup specially to handle MacOS wheels and the changes needed to support other platforms would be equivalent to writing and entire new tool, which has been done already as stated.

PR #201, opened 2024-02-06
Tweaks for Windows portability

jvolkman:
This is the other half of the patch set I've been maintaining for `repairwheel`. Hopefully these changes are innocuous enough to push upstream.

### DIFF
--- a/delocate/delocating.py
+++ b/delocate/delocating.py
@@ -7,6 +7,7 @@ import functools
 import logging
 import os
 import shutil
+import stat
 import warnings
 from os.path import abspath, basename, dirname, exists, realpath, relpath
 from os.path import join as pjoin
@@ -28,6 +29,7 @@ from typing import (
 
 from .libsana import (
     _allow_all,
+    filter_system_libs,
     get_rp_stripper,
     stripped_lib_dict,
     tree_libs,
@@ -54,6 +56,12 @@ DLC_PREFIX = "/DLC/"
 
 class DelocationError(Exception):
     pass
+
+
+def posix_relpath(path: str, start: str = None) -> str:
+    """Return path relative to start using posix separators (/)."""
+    rel = relpath(path, start)
+    return Path(rel).as_posix()
 
 
 def delocate_tree_libs(
@@ -157,7 +165,7 @@ def _analyze_tree_libs(
             # @rpath, etc, at this point should never happen.
             raise DelocationError("%s was expected to be resolved." % required)
         r_ed_base = basename(required)
-        if relpath(required, rp_root_path).startswith(".."):
+        if Path(rp_root_path) not in Path(required).parents:
             # Not local, plan to copy
             if r_ed_base in copied_basenames:
                 raise DelocationError(
@@ -202,6 +210,10 @@ def _copy_required_libs(
             "Copying library %s to %s", old_path, relpath(new_path, root_path)
         )
         shutil.copy(old_path, new_path)
+        # Make copied file writeable if necessary.
+        statinfo = os.stat(new_path)
+        if not statinfo.st_mode & stat.S_IWRITE:
+            os.chmod(new_path, statinfo.st_mode | stat.S_IWRITE)
         # Delocate this file now that it is stored locally.
         needs_delocating.add(new_path)
         # Update out_lib_dict with the new file paths.
@@ -224,7 +236,7 @@ def _update_install_names(
     for required in files_to_delocate:
         # Set relative path for local library
         for requiring, orig_install_name in lib_dict[required].items():
-            req_rel = relpath(required, dirname(requiring))
+            req_rel = posix_relpath(required, dirname(requiring))
             new_install_name = "@loader_path/" + req_rel
             if orig_install_name == new_install_name:
                 logger.info(
@@ -399,10 +411,6 @@ def _copy_required(
 
 def _dylibs_only(filename: str) -> bool:
     return filename.endswith(".so") or filename.endswith(".dylib")
-
-
-def filter_system_libs(libname: str) -> bool:
-    return not (libname.startswith("/usr/lib") or libname.startswith("/System"))
 
 
 def _delocate_filter_function(
@@ -701,7 +709,7 @@ def delocate_wheel(
         ]
         _make_install_name_ids_unique(
             libraries=libraries_in_lib_path,
-            install_id_prefix=DLC_PREFIX + relpath(lib_sdir, wheel_dir),
+            install_id_prefix=DLC_PREFIX + posix_relpath(lib_sdir, wheel_dir),
         )
         rewrite_record(wheel_dir)
         if len(copied_libs) or not in_place:
--- a/delocate/delocating.py
+++ b/delocate/delocating.py
@@ -31,6 +31,7 @@ from .libsana import (
     _allow_all,
     filter_system_libs,
     get_rp_stripper,
+    is_resolved_subpath,
     stripped_lib_dict,
     tree_libs,
     tree_libs_from_directory,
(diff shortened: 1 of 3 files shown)


delocate/delocating.py
```diff
@@ -202,6 +210,10 @@ def _copy_required_libs(
             "Copying library %s to %s", old_path, relpath(new_path, root_path)
         )
         shutil.copy(old_path, new_path)
+        # Make copied file writeable if necessary.
+        statinfo = os.stat(new_path)
+        if not statinfo.st_mode & stat.S_IWRITE:
+            os.chmod(new_path, statinfo.st_mode | stat.S_IWRITE)
```
jvolkman, line 216: This can also be found in [auditwheel](https://github.com/pypa/auditwheel/blob/db7962149a3e971eba3974a58edcb2226c787f8d/src/auditwheel/repair.py#L160-L162)

delocate/libsana.py
```diff
@@ -40,7 +41,11 @@ class DependencyNotFound(Exception):
     """
 
 
-def _filter_system_libs(libname: Text) -> bool:
+def filter_system_libs(libname: Text) -> bool:
```
jvolkman, line 44: This was duplicated between here and `delocating.py`, so I removed the latter in favor of this copy.

HexDecimal:
Honestly I think the whole source should be converted to use `pathlib`'s `Path` and `PurePosixPath`.

HexDecimal:
This kind of stuff I want to revisit by adding Ruff's [PTH](https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth) rule. I don't want these outdated calls to be refactored into *other* outdated calls.

jvolkman:
Moving to `pathlib` sounds great.

I was interested to see that there are CI tests running on windows. I don't know the full history of this project but I guess there was some cross-platform consideration at some point.

HexDecimal:
> I was interested to see that there are CI tests running on windows. I don't know the full history of this project but I guess there was some cross-platform consideration at some point.

There are a few issues such as #29 and some others about using cross platform tools. I added Windows testing at least to keep track of which features were portable, but most of them rely on either MacOS tools or linkage.

delocate/libsana.py
HexDecimal: Moving `filter_system_libs` like this technically breaks the API. Were you using it in your own library?

delocate/delocating.py
```diff
@@ -56,6 +58,12 @@ class DelocationError(Exception):
     pass
 
 
+def posix_relpath(path: str, start: str = None) -> str:
+    """Return path relative to start using posix separators (/)."""
+    rel = relpath(path, start)
+    return Path(rel).as_posix()
```
HexDecimal, line 64: If the goal is cross platform portability then do not use `str` for path types in new functions.

delocate/delocating.py
```diff
@@ -157,7 +165,7 @@ def _analyze_tree_libs(
             # @rpath, etc, at this point should never happen.
             raise DelocationError("%s was expected to be resolved." % required)
         r_ed_base = basename(required)
-        if relpath(required, rp_root_path).startswith(".."):
+        if Path(rp_root_path) not in Path(required).parents:
```
HexDecimal, line 168:
```suggestion
        if not Path(required).is_relative_to(rp_root_path):
```

delocate/libsana.py
jvolkman: Oh, that's a good point. Given the circular import issues between these to files, I guess I can just revert back to the duplication.

HexDecimal: This function is really showing its age by accepting a `Text` input. I suspect it actually works with POSIX paths exclusively. Did something happen to cause this refactor?

I've very hesitant to accept a PR with `os.path.splitdrive`.

HexDecimal (changes requested):
Recommended reading: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

I don't think now is the time for a major refactor. Just fix what's absolutely needed.

delocate/delocating.py
jvolkman: `is_relative_to` is new in 3.9 and it looks like you're testing back to 3.7.

delocate/libsana.py
jvolkman: All of these changes are the result of repairing macOS wheels on a Windows host.

If I remember correctly, callers of `filter_system_libs` are inquiring about, for example, `C:\usr\lib\libSystem.B.dylib` - which obviously doesn't exist on the repairing host - but we can still consider it as a system library. So maybe the proper fix is elsewhere upstream.

delocate/delocating.py
HexDecimal: Oops. That's easy to miss when I'm not using an IDE.

Python 3.8 EOL is this year. Maybe I can justify dropping support for it? The `path in path.parents` syntax looks terrible even compared to `relpath(...).startswith("..")`.

jvolkman: I suppose one could say the syntax isn't great, but the docs do list `root in path.parents` as equivalent to `is_relative_to`.

delocate/libsana.py
HexDecimal: I'm very interested in learning what's putting `C:` into `C:\usr\lib\libSystem.B.dylib`.

jvolkman: I can try to reproduce the issue again tomorrow. Unfortunately I don't actually have a Windows machine so it's a lot of pushing and waiting for CI. :(

HexDecimal:
> I can try to reproduce the issue again tomorrow. Unfortunately I don't actually have a Windows machine so it's a lot of pushing and waiting for CI. :(

True Windows support is going to be a difficult task. It'll be far more complex than making sure paths work correctly. I suggest closing this for now and not worrying about Windows support for the foreseeable future.

HexDecimal:
I apologize, but I'd like to put any Windows specific refactoring on hold for now. At least until certain Ruff rules are enforced and older versions of Python are dropped.

HexDecimal:
You can still make a PR for refactoring needed by `repairwheel`, if necessary.

jvolkman:
> True Windows support is going to be a difficult task. It'll be far more complex than making sure paths work correctly. 

This is true, but I've already tackled the more complicated stuff in `repairwheel` by reimplementing `otool`, `install_name_tool`, and `codesign` in Python. But I'm able to more easily patch those changes in at runtime ([here](https://github.com/jvolkman/repairwheel/blob/main/src/repairwheel/macos/repair.py#L7)).

Unless you're talking about delocating Windows dlls, for which there is already [delvewheel](https://pypi.org/project/delvewheel/).

HexDecimal:
> I've already tackled the more complicated stuff in `repairwheel` by reimplementing `otool`, `install_name_tool`, and `codesign` in Python.

Then you've already finished the hardest parts. Never mind my pessimism, it sounds like you've got it handled.

The reason I closed this is because your changes were causing readability and future refactoring concerns, such as `is_resolved_subpath` not being documented as a workaround to a missing `Path.is_relative_to`. I want to avoid the code being full of `posix_relpath` and `is_resolved_subpath` style workarounds. This would likely mean dropping Python 3.8 support before continuing, and adding linter rules to enforce the use of pathlib. These updates would conflict with this PR.

I'm handing another large PR that I'd like to merge before working on this one.

I'll reopen this to keep track of it, but it's possible that you may have to start over at a later time.

delocate/libsana.py
jvolkman: Well, since you've reopened this:
> I'm very interested in learning what's putting C: into C:\usr\lib\libSystem.B.dylib.

It comes from [here](https://github.com/matthew-brett/delocate/blob/eaba43dc531d0af348cb9b949a6c3f0790d207b2/delocate/libsana.py#L642). On windows, `realpath("foo") == "C:\\foo"`, where `C:\` is the drive letter or UNC path of the current working directory. So if you chdir into `\\myserver\shared_dir`, then `realpath("foo") == "\\\\myserver\\shared_dir\\foo"`

HexDecimal: Good work. I think `realpath` should not be called on non-existing files in this case.

If I changed the return types for `search_environment_for_lib` then I'd use `Path` for existing files and `PurePosixPath` for missing files.

PR #202, opened 2024-02-07
After copying a lib to the wheel directory, set its writeable bit if necessary

jvolkman:
Some build systems (e.g., Bazel) generate read-only artifacts, and `shutil.copy` also copies these modes.

Auditwheel has the same logic [here](https://github.com/pypa/auditwheel/blob/db7962149a3e971eba3974a58edcb2226c787f8d/src/auditwheel/repair.py#L160-L162).

### DIFF
--- a/delocate/delocating.py
+++ b/delocate/delocating.py
@@ -7,6 +7,7 @@ import functools
 import logging
 import os
 import shutil
+import stat
 import warnings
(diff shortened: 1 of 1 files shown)


HexDecimal (approved):
There's a pathlib version of this code that I'd prefer over copying from Auditwheel, but that can always be resolved later.

PR #210, opened 2024-03-23
If wheel has no package directories, use lib_sdir as a suffix

skirpichev:
Currently, there is no way to customize libs dir if the wheel has no package directory.

One use case is an extension, that offers C-API (like the gmpy2 package).  With above patch you could set one to something common for all platforms.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -10,6 +10,12 @@ rules on making a good Changelog.
 
 ## [Unreleased]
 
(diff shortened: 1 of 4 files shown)


HexDecimal:
Sorry for taking so long to respond. I've rebased the PR, updated a test, and documented the feature better. I think your implementation fits well with the existing API.

Issue #174, opened 2023-01-26
it would be nice if delocate-fuse would rename its output


glyph:
**Is your feature request related to a problem? Please describe.**
`delocate-fuse` creates a `universal2` wheel (which is amazing) but gives it the wrong tag

**Describe the solution you'd like**
Do not rewrite the input in place, instead, put a wheel next to it with a modified tag name.

**Describe alternatives you've considered**
It is not clear to me what other behavior would be correct :)

**Additional context**
I'm not sure if this is a feature or a bug. It strikes me as incorrect behavior to write an x86_64 tagged wheel that has arm64 code in it, but I am filing it as an enhancement as I assume the existing behavior has some reason for existing.

matthew-brett:
I'm afraid the reason it hasn't changed is because - it started that way.    Would you be interested in changing it to give the correct tag instead, with a `--retag` optional flag to trigger that behavior (for back-compatibility)?

Doondondon:
> **Is your feature request related to a problem? Please describe.** `delocate-fuse` creates a `universal2` wheel (which is amazing) but gives it the wrong tag
> 
> **Describe the solution you'd like** Do not rewrite the input in place, instead, put a wheel next to it with a modified tag name.
> 
> **Describe alternatives you've considered** It is not clear to me what other behavior would be correct :)
> 
> **Additional context** I'm not sure if this is a feature or a bug.

PR #215, opened 2024-04-22
Retag wheels automatically when fusing

dunkmann00:
This is my attempt at a PR for #174.

This adds the ability to "retag" a universal2 fused wheel. When running delocate-fuse with this flag, it will update the filename and the dist-info file of the fused wheel to reflect that it is a universal2 wheel.

I didn't update the readme or add any tests yet. I wanted to first get feedback on whether or not this is the "retag" you had in mind.

Let me know what you think/if there is anything you would like changed!

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,6 +14,9 @@ rules on making a good Changelog.
 
 - Improved error message for when a MacOS target version is not met.
   [#211](https://github.com/matthew-brett/delocate/issues/211)
+- Retag wheels automatically when fusing to ensure the wheel name and WHEEL
+  file tag data is accurate.
+  [#215](https://github.com/matthew-brett/delocate/pull/215)
 
 ## [0.11.0] - 2024-03-22
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,8 +14,8 @@ rules on making a good Changelog.
 
 - Improved error message for when a MacOS target version is not met.
   [#211](https://github.com/matthew-brett/delocate/issues/211)
-- Retag wheels automatically when fusing to ensure the wheel name and WHEEL
-  file tag data is accurate.
+- `delocate-fuse` will now output to a new wheel with updated tags instead of
+  replacing the first wheel given.
   [#215](https://github.com/matthew-brett/delocate/pull/215)
 
 ## [0.11.0] - 2024-03-22
--- a/Changelog.md
+++ b/Changelog.md
@@ -14,8 +14,11 @@ rules on making a good Changelog.
 
 - Improved error message for when a MacOS target version is not met.
   [#211](https://github.com/matthew-brett/delocate/issues/211)
-- `delocate-fuse` will now output to a new wheel with updated tags instead of
-  replacing the first wheel given.
+- `delocate-fuse` is no longer available and will throw an error when invoked.
+  To fuse two wheels together use `delocate-merge`. `delocate-merge` does not
+  overwrite the first wheel. It creates a new wheel with an automatically
+  determined name. If the old behavior is needed (not recommended), pin the
+  version to `delocate==0.11.0`.
   [#215](https://github.com/matthew-brett/delocate/pull/215)
 
 ## [0.11.0] - 2024-03-22
--- a/delocate/cmd/delocate_fuse.py
+++ b/delocate/cmd/delocate_fuse.py
@@ -19,7 +19,7 @@ def main() -> None:  # noqa: D103
         " name. If the old behavior is needed (not recommended), pin the"
         " version to 'delocate==0.11.0'."
     )
-    return 1
+    raise SystemExit(1)
 
 
 if __name__ == "__main__":
--- a/delocate/cmd/delocate_fuse.py
+++ b/delocate/cmd/delocate_fuse.py
@@ -34,8 +34,7 @@ def main() -> None:  # noqa: D103
     wheel1, wheel2 = [Path(wheel).resolve(strict=True) for wheel in args.wheels]
     if args.wheel_dir is not None:
         out_wheel = Path(args.wheel_dir).resolve()
-        if not out_wheel.exists():
-            out_wheel.mkdir(parents=True)
+        out_wheel.mkdir(parents=True, exist_ok=True)
     else:
         out_wheel = None
     fuse_wheels(wheel1, wheel2, out_wheel)
--- a/delocate/cmd/delocate_fuse.py
+++ b/delocate/cmd/delocate_fuse.py
@@ -1,43 +1,25 @@
 #!/usr/bin/env python3
 """Fuse two (probably delocated) wheels.
 
-Writes to a new wheel with an automatically determined name by default.
+Command is no longer available. To fuse two wheels together use
+'delocate-merge'. NOTE: 'delocate-merge' does not overwrite the first wheel. It
+creates a new wheel with an automatically determined name. If the old behavior
+is needed (not recommended), pin the version to 'delocate==0.11.0'.
 """
 
 # vim: ft=python
 from __future__ import annotations
 
-from argparse import ArgumentParser
-from pathlib import Path
-
-from delocate.cmd.common import common_parser, verbosity_config
-from delocate.fuse import fuse_wheels
-
-parser = ArgumentParser(description=__doc__, parents=[common_parser])
-parser.add_argument(
-    "wheels", nargs=2, metavar="WHEEL", type=str, help="Wheels to fuse"
-)
-parser.add_argument(
-    "-w",
-    "--wheel-dir",
-    action="store",
-    type=str,
-    help="Directory to store delocated wheels"
-    " (default is to store in the same directory as the 1st WHEEL with an"
-    " automatically determined name).",
-)
-
 
 def main() -> None:  # noqa: D103
-    args = parser.parse_args()
-    verbosity_config(args)
-    wheel1, wheel2 = [Path(wheel).resolve(strict=True) for wheel in args.wheels]
-    if args.wheel_dir is not None:
-        out_wheel = Path(args.wheel_dir).resolve()
-        out_wheel.mkdir(parents=True, exist_ok=True)
-    else:
-        out_wheel = None
-    fuse_wheels(wheel1, wheel2, out_wheel)
+    print(
+        "'delocate-fuse' is no longer available. To fuse two wheels together"
+        " use 'delocate-merge'. NOTE: 'delocate-merge' does not overwrite the"
+        " first wheel. It creates a new wheel with an automatically determined"
+        " name. If the old behavior is needed (not recommended), pin the"
+        " version to 'delocate==0.11.0'."
+    )
+    return 1
 
 
 if __name__ == "__main__":
--- a/delocate/cmd/delocate_fuse.py
+++ b/delocate/cmd/delocate_fuse.py
@@ -1,15 +1,14 @@
 #!/usr/bin/env python3
 """Fuse two (probably delocated) wheels.
 
-Overwrites the first wheel in-place by default.
+Writes to a new wheel with an automatically determined name by default.
 """
 
 # vim: ft=python
 from __future__ import absolute_import, division, print_function
 
 from argparse import ArgumentParser
-from os.path import abspath, basename, expanduser
-from os.path import join as pjoin
+from pathlib import Path
 
 from delocate.cmd.common import common_parser, verbosity_config
 from delocate.fuse import fuse_wheels
@@ -24,25 +23,22 @@ parser.add_argument(
     action="store",
     type=str,
     help="Directory to store delocated wheels"
-    " (default is to overwrite 1st WHEEL input with 2nd)",
-)
-parser.add_argument(
-    "--retag",
-    action="store_true",
-    help="Retag the fused wheel. This includes updating its filename and"
-    " dist-info (Only works when fusing to make a universal2 wheel)",
+    " (default is to store in the same directory as the 1st WHEEL with an"
+    " automatically determined name).",
 )
 
 
 def main() -> None:  # noqa: D103
     args = parser.parse_args()
     verbosity_config(args)
-    wheel1, wheel2 = [abspath(expanduser(wheel)) for wheel in args.wheels]
-    if args.wheel_dir is None:
-        out_wheel = wheel1
+    wheel1, wheel2 = [Path(wheel).resolve() for wheel in args.wheels]
+    if args.wheel_dir is not None:
+        out_wheel = Path(args.wheel_dir).resolve()
+        if not out_wheel.exists():
+            out_wheel.mkdir(parents=True)
     else:
-        out_wheel = pjoin(abspath(expanduser(args.wheel_dir)), basename(wheel1))
-    fuse_wheels(wheel1, wheel2, out_wheel, retag=args.retag)
+        out_wheel = None
+    fuse_wheels(wheel1, wheel2, out_wheel)
 
 
 if __name__ == "__main__":
--- a/delocate/cmd/delocate_fuse.py
+++ b/delocate/cmd/delocate_fuse.py
@@ -26,6 +26,12 @@ parser.add_argument(
     help="Directory to store delocated wheels"
     " (default is to overwrite 1st WHEEL input with 2nd)",
 )
+parser.add_argument(
+    "--retag",
+    action="store_true",
+    help="Retag the fused wheel. This includes updating its filename and"
+    " dist-info (Only works when fusing to make a universal2 wheel)"
+)
 
 
 def main() -> None:  # noqa: D103
@@ -36,7 +42,7 @@ def main() -> None:  # noqa: D103
         out_wheel = wheel1
     else:
         out_wheel = pjoin(abspath(expanduser(args.wheel_dir)), basename(wheel1))
-    fuse_wheels(wheel1, wheel2, out_wheel)
+    fuse_wheels(wheel1, wheel2, out_wheel, retag=args.retag)
 
 
 if __name__ == "__main__":
--- a/delocate/cmd/delocate_fuse.py
+++ b/delocate/cmd/delocate_fuse.py
@@ -30,7 +30,7 @@ parser.add_argument(
     "--retag",
     action="store_true",
     help="Retag the fused wheel. This includes updating its filename and"
-    " dist-info (Only works when fusing to make a universal2 wheel)"
+    " dist-info (Only works when fusing to make a universal2 wheel)",
 )
 
 
--- a/delocate/cmd/delocate_fuse.py
+++ b/delocate/cmd/delocate_fuse.py
@@ -5,7 +5,7 @@ Writes to a new wheel with an automatically determined name by default.
 """
 
 # vim: ft=python
-from __future__ import absolute_import, division, print_function
+from __future__ import annotations
 
 from argparse import ArgumentParser
 from pathlib import Path
@@ -31,7 +31,7 @@ parser.add_argument(
 def main() -> None:  # noqa: D103
     args = parser.parse_args()
     verbosity_config(args)
-    wheel1, wheel2 = [Path(wheel).resolve() for wheel in args.wheels]
+    wheel1, wheel2 = [Path(wheel).resolve(strict=True) for wheel in args.wheels]
     if args.wheel_dir is not None:
         out_wheel = Path(args.wheel_dir).resolve()
         if not out_wheel.exists():
--- a/delocate/cmd/delocate_merge.py
+++ b/delocate/cmd/delocate_merge.py
@@ -32,11 +32,10 @@ def main() -> None:  # noqa: D103
     args = parser.parse_args()
     verbosity_config(args)
     wheel1, wheel2 = [Path(wheel).resolve(strict=True) for wheel in args.wheels]
-    if args.wheel_dir is not None:
-        out_wheel = Path(args.wheel_dir).resolve()
-        out_wheel.mkdir(parents=True, exist_ok=True)
-    else:
-        out_wheel = None
+    out_wheel = Path(
+        args.wheel_dir if args.wheel_dir is not None else wheel1.parent
+    ).resolve()
+    out_wheel.mkdir(parents=True, exist_ok=True)
     fuse_wheels(wheel1, wheel2, out_wheel)
 
 
--- a/delocate/cmd/delocate_merge.py
+++ b/delocate/cmd/delocate_merge.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+"""Fuse two (probably delocated) wheels.
+
+Writes to a new wheel with an automatically determined name by default.
+"""
+
+# vim: ft=python
+from __future__ import annotations
+
+from argparse import ArgumentParser
+from pathlib import Path
+
+from delocate.cmd.common import common_parser, verbosity_config
+from delocate.fuse import fuse_wheels
+
+parser = ArgumentParser(description=__doc__, parents=[common_parser])
+parser.add_argument(
+    "wheels", nargs=2, metavar="WHEEL", type=str, help="Wheels to fuse"
+)
+parser.add_argument(
+    "-w",
+    "--wheel-dir",
+    action="store",
+    type=str,
+    help="Directory to store delocated wheels"
+    " (default is to store in the same directory as the 1st WHEEL with an"
+    " automatically determined name).",
+)
+
+
+def main() -> None:  # noqa: D103
+    args = parser.parse_args()
+    verbosity_config(args)
+    wheel1, wheel2 = [Path(wheel).resolve(strict=True) for wheel in args.wheels]
+    if args.wheel_dir is not None:
+        out_wheel = Path(args.wheel_dir).resolve()
+        out_wheel.mkdir(parents=True, exist_ok=True)
+    else:
+        out_wheel = None
+    fuse_wheels(wheel1, wheel2, out_wheel)
+
+
+if __name__ == "__main__":
+    main()
--- a/delocate/delocating.py
+++ b/delocate/delocating.py
@@ -654,6 +654,14 @@ def _get_archs_and_version_from_wheel_name(
             raise ValueError(f"Invalid platform tag: {platform_tag.platform}")
         major, minor, arch = match.groups()
         platform_requirements[arch] = Version(f"{major}.{minor}")
+    # If we have a wheel name with arm64 and x86_64 we have to convert that to
+    # universal2
+    if set(platform_requirements.keys()) == set(("arm64", "x86_64")):
+        version = platform_requirements["arm64"]
+        if version == Version("11.0"):
+            version = platform_requirements["x86_64"]
+        platform_requirements = {"universal2": version}
+
     return platform_requirements
 
 
--- a/delocate/delocating.py
+++ b/delocate/delocating.py
@@ -45,6 +45,7 @@ from .libsana import (
     tree_libs,
     tree_libs_from_directory,
 )
+from .pkginfo import read_pkg_info, write_pkg_info
 from .tmpdirs import TemporaryDirectory
 from .tools import (
     _is_macho_file,
@@ -867,14 +868,11 @@ def _update_wheelfile(wheel_dir: Path, wheel_name: str) -> None:
     """
     platform_tag_set = parse_wheel_filename(wheel_name)[-1]
     (file_path,) = wheel_dir.glob("*.dist-info/WHEEL")
-    with file_path.open(encoding="utf-8") as f:
-        lines = f.readlines()
-    with file_path.open("w", encoding="utf-8") as f:
-        for line in lines:
-            if line.startswith("Tag:"):
-                f.write(f"Tag: {'.'.join(str(x) for x in platform_tag_set)}\n")
-            else:
-                f.write(line)
+    info = read_pkg_info(file_path)
+    del info["Tag"]
+    for tag in platform_tag_set:
+        info.add_header("Tag", str(tag))
+    write_pkg_info(file_path, info)
 
 
 def delocate_wheel(
--- a/delocate/delocating.py
+++ b/delocate/delocating.py
@@ -656,7 +656,7 @@ def _get_archs_and_version_from_wheel_name(
         platform_requirements[arch] = Version(f"{major}.{minor}")
     # If we have a wheel name with arm64 and x86_64 we have to convert that to
     # universal2
-    if set(platform_requirements.keys()) == set(("arm64", "x86_64")):
+    if platform_requirements.keys() == {"arm64", "x86_64"}:
         version = platform_requirements["arm64"]
         if version == Version("11.0"):
             version = platform_requirements["x86_64"]
--- a/delocate/fuse.py
+++ b/delocate/fuse.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import os
 import shutil
+import tempfile
 from os import PathLike
 from os.path import exists, relpath, splitext
 from os.path import join as pjoin
@@ -24,7 +25,6 @@ from pathlib import Path
 from packaging.utils import parse_wheel_filename
 
 from .delocating import _check_and_update_wheel_name, _update_wheelfile
-from .tmpdirs import InTemporaryDirectory
 from .tools import (
     chmod_perms,
     cmp_contents,
@@ -82,7 +82,11 @@ def _retag_wheel(to_wheel: Path, from_wheel: Path, to_tree: Path) -> str:
     return retag_name
 
 
-def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
+def fuse_trees(
+    to_tree: str | PathLike,
+    from_tree: str | PathLike,
+    lib_exts=(".so", ".dylib", ".a"),
+):
     """Fuse path `from_tree` into path `to_tree`.
 
     For each file in `from_tree` - check for library file extension (in
@@ -93,14 +97,14 @@ def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
 
     Parameters
     ----------
-    to_tree : str
+    to_tree : str or Path-like
         path of tree to fuse into (update into)
-    from_tree : str
+    from_tree : str or Path-like
         path of tree to fuse from (update from)
     lib_exts : sequence, optional
         filename extensions for libraries
     """
-    for from_dirpath, dirnames, filenames in os.walk(from_tree):
+    for from_dirpath, dirnames, filenames in os.walk(str(from_tree)):
         to_dirpath = pjoin(to_tree, relpath(from_dirpath, from_tree))
         # Copy any missing directories in to_path
         for dirname in tuple(dirnames):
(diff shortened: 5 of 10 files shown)


HexDecimal:
I feel that anything updating the wheel tags should reuse the `_check_and_update_wheel_name` and `_update_wheelfile` functions.

dunkmann00:
`_update_wheelfile` looks good, I'll update it to use that (although its implementation is weird in that it doesn't use `read_pkg_info` and `write_pkg_info`)

`_check_and_update_wheel_name` I'm not sure about. Looking at it, I think I'll still have to manually update the filename to end in `universal2.whl`, since that function only deals with updating the minimum platform tag, not the arch. And I don't need to do anything fancy to determine the minimum tag like `_check_and_update_wheel_name` does (like actually checking the binaries with `macholib`). So I'm not sure using that is helpful, unless I'm missing something?

Are you okay with raising the errors like I did when the two wheels are not correct/have too many tags?

HexDecimal:
> `_update_wheelfile` looks good, I'll update it to use that (although its implementation is weird in that it doesn't use `read_pkg_info` and `write_pkg_info`)

I completely missed those functions. It would be good to update the code to use them.

> `_check_and_update_wheel_name` I'm not sure about. Looking at it, I think I'll still have to manually update the filename to end in `universal2.whl`, since that function only deals with updating the minimum platform tag, not the arch. And I don't need to do anything fancy to determine the minimum tag like `_check_and_update_wheel_name` does (like actually checking the binaries with `macholib`). So I'm not sure using that is helpful, unless I'm missing something?

Since you're updating the tag already then it would be a good idea to verify that the minimum platform tag is correct. I mostly want tag renaming to behave similarly to the `wheel-delocate` script.

I doubt that `_check_and_update_wheel_name` currently knows how to handle a universal wheel. I'll need to think about this. There seems to be a lot of edge cases.

I'd like retagging to be the default behavior rather than an optional flag. Not verifying the tags has left a lot of mis-tagged wheels around.

> Are you okay with raising the errors like I did when the two wheels are not correct/have too many tags?

I'm not sure. I like the verification and the error massage but not that it's hardcoded to `universal2`. I also find it awkward to deal with a new exception class but that's not a deal breaker.

I'd prefer a generic extendable solution to expected tags:
```py
_FUSEABLE_TAGS = {
    frozenset(("x86_64", "arm64")): "universal2",
}

```

dunkmann00:
Just pushed some new commits and I think I may have been wrong about it being overly difficult! With a slight tweak to `_get_archs_and_version_from_wheel_name` I was able to use the two functions you mentioned to get the universal2 wheel retagged. This new approach is also much simpler and I believe would handle any case. I no longer needed the universal2 checks or any error handling, so I was able to remove all of that.

I left it as a flag as I wanted to double check about making it the default behavior. Since that would be a breaking change, I just wanted to confirm that thats not a problem.

HexDecimal:
> I left it as a flag as I wanted to double check about making it the default behavior. Since that would be a breaking change, I just wanted to confirm that thats not a problem.

I'd like the default behavior changed for the reasons I mentioned above. Added or changed features must be clearly marked in the changelog and anyone relying on old behavior can pin their version of Delocate. It would be nice to have @matthew-brett's opinion as well but I'm not sure if they'll be available.

HexDecimal (changes requested):

delocate/delocating.py
```diff
@@ -653,6 +654,14 @@ def _get_archs_and_version_from_wheel_name(
             raise ValueError(f"Invalid platform tag: {platform_tag.platform}")
         major, minor, arch = match.groups()
         platform_requirements[arch] = Version(f"{major}.{minor}")
+    # If we have a wheel name with arm64 and x86_64 we have to convert that to
+    # universal2
+    if set(platform_requirements.keys()) == set(("arm64", "x86_64")):
```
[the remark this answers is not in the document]
dunkmann00: I didn't know this. Thanks for the info!

delocate/fuse.py
```diff
@@ -83,7 +123,7 @@ def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
                 _copyfile(from_path, to_path)
 
 
-def fuse_wheels(to_wheel, from_wheel, out_wheel):
+def fuse_wheels(to_wheel, from_wheel, out_wheel, retag):
```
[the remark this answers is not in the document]
dunkmann00: I agree, that is a nicer syntax.

delocate/cmd/delocate_fuse.py
```diff
@@ -1,15 +1,14 @@
 #!/usr/bin/env python3
 """Fuse two (probably delocated) wheels.
 
-Overwrites the first wheel in-place by default.
+Writes to a new wheel with an automatically determined name by default.
 """
 
 # vim: ft=python
 from __future__ import absolute_import, division, print_function
```
HexDecimal, line 8: Update `__future__` for Python 3 code. Fixes Mypy errors.
```suggestion
from __future__ import annotations
```

HexDecimal: I marked this on the wrong file. `from __future__ import annotations` needs to be added to `delocate/fuse.py`.

delocate/fuse.py
```diff
@@ -83,24 +124,46 @@ def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
...
+
+    Returns
+    -------
+    out_wheel : Path
+        The path of the new wheel from fusion of `to_wheel` and `from_wheel`.
     """
-    to_wheel, from_wheel, out_wheel = [
-        abspath(w) for w in (to_wheel, from_wheel, out_wheel)
-    ]
+    to_wheel, from_wheel = [Path(w).resolve() for w in (to_wheel, from_wheel)]
```
HexDecimal, line 153: Add `strict=True` to `Path.resolve` to validate that input paths exist.
```suggestion
    to_wheel, from_wheel = [Path(w).resolve(strict=True) for w in (to_wheel, from_wheel)]
```

delocate/cmd/delocate_fuse.py
```diff
@@ -24,18 +23,21 @@
...
 )
 
 
 def main() -> None:  # noqa: D103
     args = parser.parse_args()
     verbosity_config(args)
-    wheel1, wheel2 = [abspath(expanduser(wheel)) for wheel in args.wheels]
-    if args.wheel_dir is None:
-        out_wheel = wheel1
+    wheel1, wheel2 = [Path(wheel).resolve() for wheel in args.wheels]
```
HexDecimal, line 34:
```suggestion
    wheel1, wheel2 = [Path(wheel).resolve(strict=True) for wheel in args.wheels]
```

delocate/fuse.py
```diff
@@ -39,6 +44,42 @@ def _copyfile(in_fname, out_fname):
...
+        The path of the wheel to fuse from.
+    to_tree : Path
+        The path of the directory tree to fuse into (update into).
+
+    Returns
+    -------
+    retag_name : str
+        The new, retagged name the out wheel should be.
+    """
+    to_tree = to_tree.resolve()
```
HexDecimal, line 64: Does `to_tree` need to be resolved here? It seems it's mainly passed to other functions from here which should handle validation on their own.

HexDecimal (changes requested):
Make sure to note the changed behavior in the changelog.

delocate/fuse.py
dunkmann00: I'm not sure? Strictly speaking I guess no, a relative path here would work fine, thats what it was originally. But since `to_wheel` and `from_wheel` were resolved I thought it should be resolved at some point. Also, I don't think the other functions that use this do resolve it themselves (i.e. `_check_and_update_wheel_name` and `_update_wheelfile`).

Something else I had considered was using path instead of str in `fuse_wheels` for the trees (`"to_wheel", "from_wheel"`. Then I could just resolve them on creation and pass them in to all the functions in the `InTemporaryDirectory` context. This way it we would be passing the object around rather than rewriting the string each time. But I didn't do it since some of those functions don't accept os.PathLike. I could update those functions as well though (or convert the Path object to a string and then pass them in...but I don't love that idea).

delocate/fuse.py
```diff
@@ -83,24 +124,46 @@ def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
...
-    to_wheel, from_wheel, out_wheel = [
-        abspath(w) for w in (to_wheel, from_wheel, out_wheel)
-    ]
+    to_wheel, from_wheel = [Path(w).resolve() for w in (to_wheel, from_wheel)]
+    out_wheel = (
+        to_wheel.parent if out_wheel is None else Path(out_wheel).resolve()
+    )
+
     with InTemporaryDirectory():
         zip2dir(to_wheel, "to_wheel")
```
dunkmann00, line 159: This is what I was referring to [here](https://github.com/matthew-brett/delocate/pull/215/files#r1582361970)

```suggestion
        to_tree, from_tree = [Path(w).resolve() for w in ("to_wheel", "from_wheel")]
        zip2dir(to_wheel, to_tree)
```
And then reuse `to_tree` and `from_tree` for the rest of the function calls as well.

What do you think about that @HexDecimal?

delocate/cmd/delocate_fuse.py
dunkmann00: I'll add it to both.

delocate/fuse.py
HexDecimal: Without `strict=True` there'd be no real reason to resolve `to_wheel`/`from_wheel` either, and even that can be handled by the functions farther down. Normalizing the paths is only needed for a few specific things, and any functions which require their paths normalized should do it themselves rather than rely on the caller.

Some confusion might come from `InTemporaryDirectory` abusing the working directory. I'd personally rather have the relative paths be more strongly associated with the temporary directory. Or at least I don't immediately realize `"to_wheel"` is supposed to be a realtive path whenever I see it.
```py
with tempfile.TemporaryDirectory() as temp_dir:
    to_wheel_dir = Path(temp_dir, "to_wheel")
    from_wheel_dir = Path(temp_dir, "from_wheel")
```

Feel free to add `os.PathLike` support to any function needed. If the body is too complex to refactor then you could always do `str(Path(param))`.

Changelog.md
```diff
@@ -14,6 +14,9 @@ rules on making a good Changelog.
 
 - Improved error message for when a MacOS target version is not met.
   [#211](https://github.com/matthew-brett/delocate/issues/211)
+- Retag wheels automatically when fusing to ensure the wheel name and WHEEL
+  file tag data is accurate.
+  [#215](https://github.com/matthew-brett/delocate/pull/215)
```
HexDecimal, line 19: The changelog should mention specifically that the previous behavior of `delocate-fuse` no longer applies.
```suggestion
  `delocate-fuse` will now output to a new wheel with updated tags instead of replacing the first wheel given.
  [#215](https://github.com/matthew-brett/delocate/pull/215)
```
The readme's "Making dual-architecture binaries" section also needs to be updated. Luckily this will mostly mean deleting the documented workarounds now that the command works more intuitively.

I'm not sure how much to focus on documenting the change in the `fuse_wheels` function. It feels awkward to track changes for `0.x.x` versions. Normally you'd append a note to the docstring like this:
```py
"""
...

.. versionchanged:: 0.12
    `out_wheel` can now take a directory or None.
"""
```

matthew-brett:
Am I right in thinkng this is going to break anyone's automated workflow that using using `delocate-fuse` currently?  Because the output wheel name is going to change?

I suspect use of `delocate-fuse` is somewhat common in CI - is that a reasonable guess? 

If so - is there anything we can do not to have a torrent of 'you broke my CI' complaints?      How about (sorry if I missed it), a `--retag` command line argument, with a `--no-retag` option giving the prior default, keeping the current default, and warning that the default will change in a future version?

delocate/fuse.py
dunkmann00: Thanks for explaining all that. I sometimes, just in general not only in delocate, am unsure if I should be using relative or resolved paths, when either will work. But I think the strategy, only when it is needed, makes sense.

> I'd personally rather have the relative paths be more strongly associated with the temporary directory.

So if I'm understanding what you said correctly, you are okay with using `tempfile.TemporaryDirectory()` rather than something from `delocate.tempdirs`?

HexDecimal:
Keep in mind that this situation is a result of `delocate-fuse` being deployed in such a broken state in the first place.

> Am I right in thinkng this is going to break anyone's automated workflow that is using `delocate-fuse` currently? Because the output wheel name is going to change?

If they follow Semantic Versioning then they're fine. Anyone using the typical dependency specifiers to keep things stable, such as `delocate==0.11.0` or `delocate~=0.11.0` won't have their scripts broken. Many don't pin their dependencies though.

It's possible for those still using the workaround from the readme's "Making dual-architecture binaries" section to rename the wrong wheel after an update if they decided to not use the `--wheel-dir` flag in their script. This is the worst case scenario since the issue will be subtle, but at least the workaround suggests and demonstrates using `--wheel-dir`.

> I suspect use of `delocate-fuse` is somewhat common in CI - is that a reasonable guess?

Not nearly as common as `delocate-wheel`, but it's enough to see devs making feature requests for `delocate-fuse` on the issues tab, specifically #174 which was asking for this breaking change in the first place.

In my experience my wheels already have universal platform tags and universal dependencies before I even run Delocate on them so anything involving `delocate-fuse` would be more complex than plain `delocate-wheel`. I suspect this is the same for most devs since it's the most intuitive way to use Delocate.

> If so - is there anything we can do not to have a torrent of 'you broke my CI' complaints?

These complaints have simple answers. Either they pin version 0.11.0 or they modify their scripts to handle the updated output. If they read the changelog then they'll know everything they need without having to raise an issue.

I think that sorting out complaints would take less time and effort then dealing with the technical debt of keeping the old behavior in the current code.

There were CI complaints for 0.11.0 as well and those weren't hard to deal with.

> How about (sorry if I missed it), a `--retag` command line argument, with a `--no-retag` option giving the prior default, keeping the current default, and warning that the default will change in a future version?

They can pin `delocate==0.11.0` to invoke the old behavior is my proposal.

It's mainly you who wants to keep the old behavior as an option. The main issue solved is that the old behavior is unintuitive especially due to how `delocate-fuse` never even asks for an output name and simply outputs into one of the inputs. Having to manually rename the wheel afterwards is also prone to issues and typically results in a mis-tagged wheel.

delocate/fuse.py
HexDecimal: > So if I'm understanding what you said correctly, you are okay with using `tempfile.TemporaryDirectory()` rather than something from `delocate.tempdirs`?

Yes.

Decloate used to be a Python 2 package, so it's full of backports of features that are now native to Python 3. This means the stuff in `delocate.tempdirs` is outdated.

Changelog.md
dunkmann00: - [x] Changelog
- [x] README
- [ ] Docstring, I updated it the way I think I was supposed to. But you can double check this and see if its to your liking.

delocate/fuse.py
```diff
@@ -50,14 +97,14 @@ def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
...
+    to_tree : str or Path-like
         path of tree to fuse into (update into)
-    from_tree : str
+    from_tree : str or Path-like
         path of tree to fuse from (update from)
     lib_exts : sequence, optional
         filename extensions for libraries
     """
-    for from_dirpath, dirnames, filenames in os.walk(from_tree):
+    for from_dirpath, dirnames, filenames in os.walk(str(from_tree)):
```
HexDecimal, line 107: `str` works for `str | Path` but `PathLike` only means the type can convert to `Path` and not that its `__str__` returns a valid path. You always have to convert to `Path` before converting to `str`.

Python standard library functions can take `str | PathLike` types directly.
```suggestion
    for from_dirpath, dirnames, filenames in os.walk(from_tree):
```

delocate/fuse.py
```diff
@@ -83,24 +130,50 @@ def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
...
+        The path of the new wheel from fusion of `to_wheel` and `from_wheel`.
+
+    .. versionchanged:: 0.12
+        `out_wheel` can now take a directory or None.
     """
-    to_wheel, from_wheel, out_wheel = [
-        abspath(w) for w in (to_wheel, from_wheel, out_wheel)
+    to_wheel, from_wheel = [
+        Path(w).resolve(strict=True) for w in (to_wheel, from_wheel)
     ]
```
HexDecimal, line 164: This probably doesn't need to be a comprehension at this point.
```suggestion
    to_wheel = Path(to_wheel).resolve(strict=True)
    from_wheel = Path(from_wheel).resolve(strict=True)
```

delocate/wheeltools.py
```diff
@@ -34,7 +35,7 @@ def _open_for_csv(name, mode):
     return open_rw(name, mode, newline="", encoding="utf-8")
 
 
-def rewrite_record(bdist_dir: str) -> None:
+def rewrite_record(bdist_dir: str | PathLike) -> None:
```
HexDecimal, line 38: Needs `from __future__ import annotations` to resolve this Mypy error.

HexDecimal (changes requested):

Changelog.md
HexDecimal: The docstring looks okay.

delocate/fuse.py
dunkmann00: > Python standard library functions can take str | PathLike types directly.

Thats what I thought, but mypy was giving me an error when I ran it locally. I'll push the change to see what happens on the CI.

HexDecimal: I see the error. Mypy thinks PathLike for `os.walk` is supported since Python 3.8, but the Python docs say it's supported since Python 3.6.

dunkmann00: Mypy is okay with:

```python
for from_dirpath, dirnames, filenames in os.walk(Path(from_tree)):
```

Which is still a pathlike, but for some reason doesn't error. I know `pathlib` path's have their own glob method, but it has slightly different behavior with symlinks. So if this is okay, I can just wrap `from_tree` in `Path` and leave it in `os.walk`.

HexDecimal: That's good if that works. I wasn't able to figure out the exact cause of the error, but Python 3.7 is end-of-life and will probably have more of these kinds of issues until it's dropped.

dunkmann00: Just found this in `dir2zip`:

```python
def dir2zip(
    in_dir: str | PathLike[str],
    zip_fname: str | PathLike[str],
    *,
    compression: int = zipfile.ZIP_DEFLATED,
    compress_level: int = -1,
    date_time: Optional[_DateTuple] = None,
) -> None:
```

So maybe just change the type to `str | PathLike[str]`?

dunkmann00: I went with wrapping it in Path approach.

delocate/cmd/delocate_fuse.py
```diff
@@ -24,18 +23,21 @@
...
     args = parser.parse_args()
     verbosity_config(args)
-    wheel1, wheel2 = [abspath(expanduser(wheel)) for wheel in args.wheels]
-    if args.wheel_dir is None:
-        out_wheel = wheel1
+    wheel1, wheel2 = [Path(wheel).resolve(strict=True) for wheel in args.wheels]
+    if args.wheel_dir is not None:
+        out_wheel = Path(args.wheel_dir).resolve()
+        if not out_wheel.exists():
+            out_wheel.mkdir(parents=True)
```
HexDecimal, line 38: This can be simplified.
```suggestion
        out_wheel.mkdir(parents=True, exist_ok=True)
```

delocate/delocating.py
```diff
@@ -653,6 +654,14 @@ def _get_archs_and_version_from_wheel_name(
             raise ValueError(f"Invalid platform tag: {platform_tag.platform}")
         major, minor, arch = match.groups()
         platform_requirements[arch] = Version(f"{major}.{minor}")
+    # If we have a wheel name with arm64 and x86_64 we have to convert that to
+    # universal2
+    if platform_requirements.keys() == {"arm64", "x86_64"}:
+        version = platform_requirements["arm64"]
+        if version == Version("11.0"):
+            version = platform_requirements["x86_64"]
+        platform_requirements = {"universal2": version}
```
HexDecimal, line 663: This branch is not covered by tests at the moment. I assumed that delocating a universal wheel would meet this condition but it seems I was wrong.

HexDecimal (changes requested):

matthew-brett:
> It's mainly you who wants to keep the old behavior as an option.

Well - it's not me of course - it's me speculating about our users.

In my experience, the only people who pay any attention to Semantic Versioning are the authors of the packages doing the version bump - it was a nice idea, but I bet you'll find it hard to find any CI scripts that pay any attention to major-version bumps.

Another option would be to rename the command to `delocate-fuse-universal` and have `delocate-fuse` raise an error pointing to `delocate -fuse-universal`.    At least that way no-one will be confused when they get the error.

HexDecimal:
> Another option would be to rename the command to `delocate-fuse-universal` and have `delocate-fuse` raise an error pointing to `delocate -fuse-universal`. At least that way no-one will be confused when they get the error.

I've considered replacing the syntax of `delocate-fuse` to force users to notice the update. It isn't a solution I like, but it's the only option effective at preventing the most subtle issues from ever happening.

Something that would work for most cases would be to make `--wheel-dir` a required flag. At least temporarily.

matthew-brett:
Anything that raises an informative error would be fine I think.

dunkmann00:
What if we add a flag `--overwrite/--no-overwrite`. The default when neither is given, **for now**, being `--overwrite`. But in this case, we raise the informative error @matthew-brett would like to see to notify users they must use the `--no-overwrite` flag in versions `0.12` or over. In a future version we can change the default to `--no-overwrite` or just remove the flag entirely.

So as an example:
```bash
delocate-fuse wheel1.whl wheel2.whl  # Raises error where we inform the need for `--no-overwrite`

delocate-fuse --no-overwrite wheel1.whl wheel2.whl  # Runs without error
```

In this case, we probably would not need to raise any error if `--wheel-dir` was used. So maybe thinking about the error message, it could be something that indicates for version `0.12` and above, overwriting the first wheel is an error, you must use either `--no-overwrite` or `--wheel-dir`, and in a future version `--no-overwrite` will become the default behavior.

delocate/delocating.py
dunkmann00: I've added a test for this case in the `test_get_archs_and_version_from_wheel_name` test. Delocating a universal2 wheel doesn't enter this path because the tag from the name is already universal2, not one tag with arm64 and another with x86_64, which is what this fixes.

dunkmann00:
I just fixed the last test I added. Total user error, forgot to add the right hand side of that assert 😅 sorry about that.

HexDecimal:
> What if we add a flag `--overwrite/--no-overwrite`. The default when neither is given, **for now**, being `--overwrite`. But in this case, we raise the informative error matthew-brett would like to see to notify users they must use the `--no-overwrite` flag in versions `0.12` or over. In a future version we can change the default to `--no-overwrite` or just remove the flag entirely.

This would sound perfectly reasonable, but the issue with this is that no matter how loudly you try to announce the change, devs won't notice it until the change is finalized and their scripts break.

The reason I suggest requiring `--wheel-dir` is that this would break almost any script relying on the wheel reusing the name of the existing wheels. Those scripts would fail to find the wheel with the old name in the given directory and crash. **Scripts not using `--wheel-dir` will never notice that there's a new wheel and will always use the incorrect wheel for the rest of the script which is why they have to be broken deliberately.** With `--wheel-dir` the wheel outputted to the wheel directory is always the fused wheel even when a broad wildcard is used to select it. The only bad case is when `--wheel-dir` is the same directory as where the inputs are, but that is not how the flag is typically used.

For announcing the changes to the user, I think the most reasonable thing you can do is simply print where wheel was outputted to stdout. By adding the following to the end of the `delocate-fuse` main function:
```py
print(f"Fused wheel written to {out_wheel}")
```
If the user doesn't notice this then they're not going to notice anything else less than an error.

dunkmann00:
I'm suggesting we don't just print something out, but raise an error and print something out. I thought based on @matthew-brett's comments:
> Another option would be to rename the command to delocate-fuse-universal and have delocate-fuse raise an error pointing to delocate -fuse-universal
...
Anything that raises an informative error would be fine I think.

And what I believe your position is, which is we should retag by default, not because it is a good idea, but because doing anything else is incorrect and leads to bad wheels out in the wild, when they could have easily been made correct.

That this satisfies both of those interests.

So what I was suggesting would only work if the user specifically adds `--no-overwrite`, in which case it would retag the wheel and write it to the new output file. Or if the user has the `--wheel-dir` flag, because for the reasons you just explained, we can write the retagged wheel correctly, and the way a script would find it is most likely unaffected, so no need to let the user know of anything in that case.

But if/when this upgrade is made, and those who have disregarded semantic versioning run their CI the same as before, we can now raise an error explicitly stating they now need to use the `--no-overwrite` flag or `--wheel-dir` argument.

Perhaps I was unclear because I included the `--overwrite` option, in my last message. I just put that because I thought that is how `argparse` does flags, maybe I am getting it mixed up with `click`. But the point would be `--overwrite` wouldn't run successfully. If no flag is passed in, it would get treated as if `--overwrite` was given on the command line. But this would raise the error where we point them to this new change and what they can do to migrate to the new version. Breaking their script, getting their attention, but because of the useful output from the error message, avoiding a bunch of support issues.

So I think this is a reasonable middle ground where we force the user to change, let them know how they can do that, but don't really add any substantial extra code thats going to lay around in the library and be hard to remove in the future.

matthew-brett:
Just to clarify - that was my intention with `delocate-fuse-universal` - that you just strip `delocate-fuse` back to an informative error, and don't keep the previous behavior anywhere as a maintenance burden.

HexDecimal:
> And what I believe your position is, which is we should retag by default, not because it is a good idea, but because doing anything else is incorrect and leads to bad wheels out in the wild, when they could have easily been made correct.

That's correct. This is why I insist on the new behavior being default and removing even the option (other when pinning an old version of delocate) for the previous behavior. I couldn't come up with anything better than making `--wheel-dir` a requirement since anything else would make the 'correct' behavior of fixing the platform tags not the default.

`--overwrite` is also too ambiguous a name for this flag, the word 'overwrite' can mean too many things outside of what this actually does. It also preserves the old behavior which I obviously dislike.

`delocate-fuse-universal` is better. At the risk of bike-shedding, we could also use a different word with the same meaning such as `delocate-splice`. Just keep in mind that the new name is permanent. The changelog needs to be clear that the name has changed. Then `delocate-fuse` prints out the informative error and returns a non-zero exit code no matter what arguments are given. The readme would be updated to use the new name.

I insist on pinning `delocate==0.11.0` to be the suggested option for preserving the old behavior. This should be mentioned in the changelog and the informative error.

Although I feel strongly about this, I'll go with whatever solution Matthew asserts.

matthew-brett:
I'm happy with whatever name for the new command - and it's perfectly reasonable to not want to keep the old pathway around.  As long as the error message is clear, and there is no possibility of unexpected behavior without the error, anything is fine with me.

dunkmann00:
@HexDecimal I don't understand why you feel a new command is better than requiring a new flag on `delocate-fuse`. I don't have any preference on the name but something like:
```python
# delocate_fuse.py

...

parser.add_argument(
    "--no-overwrite",
    action="store_true"
)

def main() -> None:  # noqa: D103
    args = parser.parse_args()
    if not (args.no_overwrite or args.wheel_dir):
        raise RuntimeError("Useful error message telling user to use '--no-overwrite' or '--wheel-dir'") # This returns 1 as an exit code
    
    ...

```

This causes the same kind of exposure to the new change as setting up a new command like `delocate-fuse-wheel`, but has two benefits I can think of:
1) We don't have to bother people already using `--wheel-dir`
2) It can be temporary. In the future `--no-overwrite` becomes default:
```python
parser.add_argument(
    "--no-overwrite",
    action="store_true",
    default=True
)
```

This way its no longer necessary to use and we can go back to telling people to do `delocate-fuse wheel1.whl wheel2.whl` the way you'd prefer to, but for those who at that point in time have it in their script, it won't cause any unrecognized argument errors so it would be a painless transition.

Another more subjective benefit is we don't have to permanently change the name and lose `delocate-fuse` which to me is short, sweet, and to the point.

Again if you feel `--no-overwrite` is a poor name, thats fine, but what I would be more happy with is if you understood the idea I'm trying to get across. If you still feel going with a new command is better, than thats fine as well. But I'm not sure you have fully understood what I'm trying to suggest.

HexDecimal:
> I don't understand why you feel a new command is better than requiring a new flag on `delocate-fuse`. I don't have any preference on the name but something like:

It's a good attempt, but I see downsides to this approach.

> It can be temporary. In the future `--no-overwrite` becomes default:

Once `--no-overwrite` becomes the default then that flag becomes a no-op. You then have a flag that does nothing and has to be left in for backwards compatibility since removing it later means a 2nd break in the command API. Renaming the command doesn't have this problem.

Rationally I should be okay with leaving `--no-overwrite` in forever, but I'm too obsessive compulsive to quietly accept that.

For names, the original name of `--retag` was good enough, or something like `--fix-tags`/`--update-tags`, but again I don't want to leave these in once they're the default.

> We don't have to bother people already using `--wheel-dir`

Those using `--wheel-dir` are not completely in the clear. Their scripts can still break without an explanation of why. The change in command name will guarantee that all users know exactly what has changed. Changing the command name does the best job of handling Matthew's worry that users will be confused by the change in behavior.

> Another more subjective benefit is we don't have to permanently change the name and lose `delocate-fuse` which to me is short, sweet, and to the point.

Yeah, I got used to the name too, but its API has made it difficult to fix in a graceful way. I'll let you pick the new name if you want.

`delocate-fuse-universal` is too long. No other commands are three words long and sometimes it feels like technically the command might have utility outside of universal wheels, assuming combined architectures are still called universal in the future. Could've easily been named `delocate-fuse-fat` with this methodology.

`delocate-splice` is pretty clear wordplay, but it might be too edgy or fancy.

`delocate-merge` is clear and is well recognized for merging items together. Also `delocate-combine` as another generic name.

dunkmann00:
Sorry I wasn't able to respond to this sooner. I wanted to say I very much appreciate that response @HexDecimal and completely agree with all of your critiques. With that, I've put together the changes necessary to move all the functionality of fusing into `delocate-merge` (I liked that one the best 👍).

Let me know what you think.

delocate/cmd/delocate_fuse.py
```diff
@@ -1,42 +1,25 @@
...
-        out_wheel = pjoin(abspath(expanduser(args.wheel_dir)), basename(wheel1))
-    fuse_wheels(wheel1, wheel2, out_wheel)
+    print(
+        "'delocate-fuse' is no longer available. To fuse two wheels together"
+        " use 'delocate-merge'. NOTE: 'delocate-merge' does not overwrite the"
+        " first wheel. It creates a new wheel with an automatically determined"
+        " name. If the old behavior is needed (not recommended), pin the"
+        " version to 'delocate==0.11.0'."
+    )
+    return 1
```
HexDecimal, line 22: The return value from `main` isn't used by the caller. Simplest solution is to raise `SystemExit`.
```suggestion
    raise SystemExit(1)
```

dunkmann00: But this is what the entry point script looks like:
```python
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```
The return value from main is used as the exit code. `sys.exit` raises `SystemExit`.

HexDecimal: You're right. I was looking at the scripts own call of `main`, not the entry point code.

The command scripts themselves are not set up correctly to return `int` unless they're modified. Raising `SystemExit` does the same without any other changes.

delocate/cmd/delocate_merge.py
```diff
@@ -0,0 +1,44 @@
...
+
+def main() -> None:  # noqa: D103
+    args = parser.parse_args()
+    verbosity_config(args)
+    wheel1, wheel2 = [Path(wheel).resolve(strict=True) for wheel in args.wheels]
+    if args.wheel_dir is not None:
+        out_wheel = Path(args.wheel_dir).resolve()
+        out_wheel.mkdir(parents=True, exist_ok=True)
+    else:
+        out_wheel = None
```
HexDecimal, line 39: Getting the first wheel directory can be handled here instead of the `fuse_wheel` function.
```suggestion
    out_wheel = Path(
        args.wheel_dir if args.wheel_dir is not None else wheel1.parent
    ).resolve()
    out_wheel.mkdir(parents=True, exist_ok=True)
```

HexDecimal (changes requested):
Lots of progress made so far, I think anything left is polish and cleanup. I like the warning added to the readme.

No need to rush. I'd prefer this to be done correctly rather than quickly.

delocate/fuse.py
```diff
@@ -83,24 +130,49 @@ def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
                 _copyfile(from_path, to_path)
 
 
-def fuse_wheels(to_wheel, from_wheel, out_wheel):
+def fuse_wheels(
+    to_wheel: str | PathLike,
+    from_wheel: str | PathLike,
+    out_wheel: str | PathLike | None = None,
```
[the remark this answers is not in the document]
dunkmann00: If I could just throw something out there. When I used the `delocate` library to fuse wheels in a python script, rather than running the command, I just called `fuse_wheels` directly. Since I was already running python, this worked great and was easier than having to use subprocess.

With `delocate-merge` I could do the same thing. But if we move the `out_wheel` logic into the command function, it becomes a little less convenient.

I'm not sure if that is a use case you care to support, no big deal if not. But I figured I'd bring it up to see what you think.

dunkmann00:
Just a heads up, I'm not going to have much time to work on this for the next week. So I'll do what I can, but probably will get to the remaining things towards the end of next week.

delocate/fuse.py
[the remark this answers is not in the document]
HexDecimal: This suggestion was to reduce complexity of the function before the change becomes permanent, "explicit is better than implicit" and all that, so that there are fewer ambiguous contexts.

For example: Are you sure you wanted your script to use the input wheels directory for the output wheel? And not something like a temporary directory? Is the best default directory for a CLI context and API context the same directory?

I'd rather remove the default here rather than commit to an answer. You can add `out_wheel=to_wheel.parent` to replicate the current behavior in your script, it's more explicit and easier for anyone reading it to understand, and any future updates are less likely to break it.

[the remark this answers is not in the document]
dunkmann00: > Is the best default directory for a CLI context and API context the same directory?

Interesting way of putting it. I wasn't thinking of it that way.

And yes, your suggestion makes sense 👍

dunkmann00:
Any thoughts on what's left for this to be ready to merge?

HexDecimal (approved):
As far as I can tell. Everything is resolved.

This PR can be squashed and merged.

dunkmann00:
Great! Thanks for working with me through this.

glyph:
Awesome to have this fixed! I will have to look at fixing Encrust to make sure I'm not expecting the old name :)

glyph:
And, done: https://pypi.org/project/encrust/2024.9.3/

PR #216, opened 2024-05-11
fix string formatting in error message

jameslamb:
Thanks for this awesome project!

I noticed tonight that an error message is printing with the literal string `f{lib_path}` in it, like this:

> delocate.libsana.DelocationError: f{lib_path} already exists in wheel but need to copy /opt/homebrew/Cellar/libomp/17.0.6/lib/libomp.dylib

I strongly suspect that was supposed to be an f-string, where `{lib_path}` is replaced in the error message with a path to a library.

This proposes that fix.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -21,6 +21,12 @@ rules on making a good Changelog.
   version to `delocate==0.11.0`.
   [#215](https://github.com/matthew-brett/delocate/pull/215)
 
+### Fixed
+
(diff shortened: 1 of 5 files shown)


HexDecimal:
I apologize for getting around to this PR so late. I've rebased the PR and updated the changelog and the relevant tests. Tests for `DelocationError` were missing a `match=` argument.

jameslamb:
No problem, and thanks very much!

PR #219, opened 2024-07-06
Fix platform tag version regression for macOS 11 and later

### LINKED ISSUE #214: Tool `delocate-wheel` alters tag in wheel name to seemingly invalid tag/number
**Describe the bug**
Tool `delocate-wheel` alters tag in wheel name to seemingly invalid tag/number
We are using GitHub-runners to generate som large macOS wheels for repository `Carolina`

**To Reproduce**

The wheel before we use `delocate`:
```
  Created wheel for carolina: filename=carolina-1.0.16.dev3+g42cf9bc-cp38-cp38-macosx_11_0_x86_64.whl size=32980 sha256=f719c488eded81a29c5685e636b6b290f516d8fab9cbfde7b626041e49240904
...
```

Running `delocate-wheel`:
```
delocate-wheel -w /tmp/carolina_dist -v $unfixed_wheel_path
...
...
-rw-r--r--  1 runner  wheel    17M Apr  5 13:42 carolina-1.0.16.dev3+g42cf9bc-cp38-cp38-macosx_13_6_x86_64.whl
```

I see that there are dependencies that would indicate this to be set to `13_x`, but as far as I can tell the wheel won't work when `13_6` is used. If I rename the file I can install it locally. The same is true for `11_0`.

```
Compatible tags: 1835
  cp38-cp38-macosx_13_0_x86_64
  cp38-cp38-macosx_13_0_intel
  cp38-cp38-macosx_13_0_fat64
  cp38-cp38-macosx_13_0_fat32
  cp38-cp38-macosx_13_0_universal2
  cp38-cp38-macosx_13_0_universal
  ...
```

**Wheels used/generated**

```
# generated using version 0.10.7 -- works as intended
carolina-1.0.16.dev2+g9c153fd-cp311-cp311-macosx_11_0_x86_64.whl

# generated using version 0.11.0 -- works when renaming to `13_0` (or `11_0`)
carolina-1.0.16.dev3+g42cf9bc-cp311-cp311-macosx_13_6_x86_64.whl
```

[carolina-1.0.16.dev3+g42cf9bc-cp311-cp311-macosx_13_6_x86_64.whl.tar.gz](https://github.com/matthew-brett/delocate/files/14904281/carolina-1.0.16.dev3%2Bg42cf9bc-cp311-cp311-macosx_13_6_x86_64.whl.tar.gz)

[carolina-1.0.16.dev2+g9c153fd-cp311-cp311-macosx_11_0_x86_64.whl.tar.gz](https://github.com/matthew-brett/delocate/files/14904284/carolina-1.0.16.dev2%2Bg9c153fd-cp311-cp311-macosx_11_0_x86_64.whl.tar.gz)

 - OS versions macOS 13 and macOS 14
 - Delocate version: 0.11.0

**Additional context**
Works with previous release, 0.10.7

HexDecimal:
Zero out minor version of platform tags for macOS 11 and later which fixes the invalid tags for those releases. Closes #214 

Delocate has a bias for testing older versions of macOS. I've added a test to ensure this works as intended.

Also discussed about in #214. This updates the behavior of `MACOSX_DEPLOYMENT_TARGET` and `--require-target-macos-version`. These will now raise the resulting platform tag versions if they are lower than the requested version. I'm certain this is desirable for `MACOSX_DEPLOYMENT_TARGET` but I'm not sure if it should be the same for `--require-target-macos-version`.


HexDecimal:
Multiple sections of code had mislabeled `intel` as `universal`. I didn't see any issues on this, possibly because those platforms are old. These have been fixed.

Only `universal2` and `intel` are supported as combined architectures right now. In theory more can be supported but it's hard to find documentation on which architectures map to each other. The most I found for this would be [in the packaging source code](https://github.com/pypa/packaging/blob/793ee28b4c70886e8de5731e24e444388916b3ee/src/packaging/tags.py#L366-L395). For now `universal2` seems to be the only one being used publicly.

I've decided to deprecate `--require-target-macos-version` since it causes confusion with `MACOSX_DEPLOYMENT_TARGET`. From now on all instructions should only mention `MACOSX_DEPLOYMENT_TARGET`.

PR #220, opened 2024-07-06
Configure trusted publishing

### LINKED ISSUE #196: [maintenance hint] Use `pypi-publish` action with secretless publishing from GHA
I was skimming through the recent changes as I was trying to figure out a new regression and noticed something that should probably be fixed — the GHA workflow calls Twine directly and uses a long-living API token or even a user-wide password (which is worse, security-wise).

There's a more secure and easier way of doing this now which my action (yes, it's a shameless plug!) has supported since the early spring, way before it's gone GA: https://github.com/marketplace/actions/pypi-publish#trusted-publishing.

My PyPUG guide is also updated with a full usage example of secretless publishing: https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/.

I suggest you upgrade the automation to be more in line with the modern practices :)

HexDecimal:
Replaces my API token with trusted publishing as described in #196. Closes #196

I have enough access to handle on rest on PyPI but a GitHub user can't setup an environment on another users private repo. @matthew-brett will need to do the following:

* On this repo, go to: Setting > Enviroments > New environment
* Name the new environment: `deploy`
* Optionally setup deployment protection rules

Only then can I finish this PR.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -185,6 +185,9 @@ jobs:
     needs: isolated_tests
     if: startsWith(github.ref, 'refs/tags/')
     runs-on: ubuntu-latest
+    environment:
+      name: deploy
+      url: https://pypi.org/project/delocate/
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
@@ -202,8 +205,6 @@ jobs:
         with:
           name: delocate-sdist
           path: dist/
-      - name: Upload to PyPI
-        env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-        run: twine upload dist/* --non-interactive --skip-existing
+      - uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          skip-existing: true


matthew-brett:
Done - anything else I need to do?

HexDecimal:
No. I can handle the rest. I can't see the new environment but I'll assume it's there.

.github/workflows/python-package.yml
```diff
@@ -185,6 +185,9 @@ jobs:
     needs: isolated_tests
     if: startsWith(github.ref, 'refs/tags/')
     runs-on: ubuntu-latest
+    environment:
+      name: deploy
+      url: https://pypi.org/project/delocate/
```
webknjaz, line 190: FYI you can also append `${{ github.ref_name}}` here for nicer URLs.

.github/workflows/python-package.yml
```diff
@@ -202,8 +205,6 @@ jobs:
...
           name: delocate-sdist
           path: dist/
-      - name: Upload to PyPI
-        env:
-          TWINE_USERNAME: __token__
-          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-        run: twine upload dist/* --non-interactive --skip-existing
+      - uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          skip-existing: true
```
webknjaz, line 210: I usually consider this feature harmful — are there multiple upload sources?

.github/workflows/python-package.yml
```diff
@@ -185,6 +185,9 @@ jobs:
     needs: isolated_tests
     if: startsWith(github.ref, 'refs/tags/')
     runs-on: ubuntu-latest
+    environment:
+      name: deploy
```
webknjaz, line 189: Technically, the deployment target is PyPI, not Deploy. And that would differ from TestPyPI. So I tend to suggest those names that correspond to things they are supposed to represent.

.github/workflows/python-package.yml
HexDecimal: No, that was just habit, and the previous script was already using `--skip-existing`, and my other projects I was referencing were also using it.

It'd be reasonable to remove it.

.github/workflows/python-package.yml
HexDecimal: That could work. I remember the first examples using the PyPI URL, but it's hard to tell what's most relevant.

.github/workflows/python-package.yml
HexDecimal: I had also thought of `PyPI`, but I didn't know if anything was standard and wanted to keep things simple since I didn't know that the environment would be created automatically anyways.

HexDecimal:
I was happy enough with this the way it was, but I can accept a PR or update this myself if anyone requests it. I'll update the configuration from the PyPI side if necessary.

.github/workflows/python-package.yml
webknjaz: Well, I wrote those examples and never added this bit. But I use it in real-life projects (though, the workflows are more complicated and I wanted to keep the tutorial simple).

.github/workflows/python-package.yml
webknjaz: You can reconfigure whenever. I use `pypi` and `testpypi` everywhere. Just remember to also update the trust config in PyPI.

.github/workflows/python-package.yml
webknjaz: Also, the reason there's `startsWith()` is that `ref_name` didn't exist at the time of writing and GH added it later.

HexDecimal: I'm going to have to update *a lot* of projects.

PR #234, opened 2024-12-17
Append delocate version to wheel metadata

HexDecimal:
Adds the delocate version used to wheel metadata as suggested by the [Binary distribution format guidelines](https://packaging.python.org/en/latest/specifications/binary-distribution-format). The format is `delocate {version}` as shown in the guideline but I've seen other tools use `name ({version})` instead.

Tagging wheels like this will help with future issues if there's ever a long term problem with a specific version of delocate.

There seemed to be cases where `rewrite_record` could be called before modifications were done, this has been resolved.

Was also able to move `.coveragerc` into `pyproject.toml` so that's one less config file.

### Pull Request Checklist

- [x] Read and follow the [CONTRIBUTING.md](https://github.com/matthew-brett/delocate/blob/master/CONTRIBUTING.md) guide
- [x] Mentioned relevant [issues](https://github.com/matthew-brett/delocate/issues)
- [x] Append public facing changes to [Changelog.md](https://github.com/matthew-brett/delocate/blob/master/Changelog.md)
- [x] Ensure new features are covered by tests
- [x] Ensure fixes are verified by tests

### DIFF
--- a/delocate/fuse.py
+++ b/delocate/fuse.py
@@ -49,7 +49,7 @@ def _copyfile(in_fname, out_fname):
 
 
 def _retag_wheel(to_wheel: Path, from_wheel: Path, to_tree: Path) -> str:
-    """Update the name and dist-info to reflect a univeral2 wheel.
+    """Update the name and dist-info to reflect a universal2 wheel.
 
     Parameters
     ----------
@@ -68,8 +68,8 @@ def _retag_wheel(to_wheel: Path, from_wheel: Path, to_tree: Path) -> str:
     to_tree = to_tree.resolve()
     # Add from_wheel platform tags onto to_wheel filename, but make sure to not
     # add a tag if it is already there
-    from_wheel_tags = parse_wheel_filename(from_wheel.name)[-1]
-    to_wheel_tags = parse_wheel_filename(to_wheel.name)[-1]
+    _, _, _, from_wheel_tags = parse_wheel_filename(from_wheel.name)
+    _, _, _, to_wheel_tags = parse_wheel_filename(to_wheel.name)
     add_platform_tags = (
         f".{tag.platform}" for tag in from_wheel_tags - to_wheel_tags
     )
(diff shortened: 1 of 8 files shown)


HexDecimal:
After some consideration I want to write this to not use `importlib.metadata`.

delocate/fuse.py
```diff
@@ -68,8 +68,8 @@ def _retag_wheel(to_wheel: Path, from_wheel: Path, to_tree: Path) -> str:
     to_tree = to_tree.resolve()
     # Add from_wheel platform tags onto to_wheel filename, but make sure to not
     # add a tag if it is already there
-    from_wheel_tags = parse_wheel_filename(from_wheel.name)[-1]
-    to_wheel_tags = parse_wheel_filename(to_wheel.name)[-1]
+    _, _, _, from_wheel_tags = parse_wheel_filename(from_wheel.name)
```
matthew-brett, line 71: How about:
```suggestion
    *_, from_wheel_tags = parse_wheel_filename(from_wheel.name)
```

delocate/fuse.py
```diff
@@ -68,8 +68,8 @@ def _retag_wheel(to_wheel: Path, from_wheel: Path, to_tree: Path) -> str:
...
+    _, _, _, to_wheel_tags = parse_wheel_filename(to_wheel.name)
```
matthew-brett, line 72: Similarly:

```suggestion
    *_, to_wheel_tags = parse_wheel_filename(to_wheel.name)
```

matthew-brett (approved):
A cosmetic suggestion, otherwise, looks good.

delocate/fuse.py
HexDecimal: I prefer the explicit syntax over tuple repacking or last item index since this clarifies the size of the tuple being unpacked and which items are being unpacked onto which names as well as making clear that these items are distinct from each other and that this isn't just a plain list with the last item being accessed.

Your suggestion is equivalent to a more complex last index access, in which case I should've just kept these lines unchanged.

A custom NamedTuple could also work here. Anything to clarify that the return value is a structured tuple.

matthew-brett:
That’s fine; choose what you prefer, just a suggestion.

PR #236, opened 2025-01-03
Smarter detection of binary files during delocate-merge

### LINKED ISSUE #228: delocate-merge does not merge Qt Framework binaries
**Describe the bug**
While delocate-merge succeeds to merge PyQt6-Qt6 arm and x86 wheels the resulting wheel contains Qt Frameworks that are not universal2.

**To Reproduce**
```
# delocate-merge PyQt6_Qt6-6.7.2-py3-none-macosx_10_14_x86_64.whl PyQt6_Qt6-6.7.2-py3-none-macosx_11_0_arm64.whl -w .
# unzip PyQt6_Qt6-6.7.2-py3-none-macosx_11_0_universal2.whl.zip
# cd PyQt6/Qt6/lib/QtCore.framework/Versions/A
# file QtCore
QtCore: Mach-O 64-bit dynamically linked shared library arm64
```
**Expected behavior**
A full universal2 wheel.

**Wheels used**
If a wheel is involved then consider attaching the original wheel (before being delocated) or linking to the repository where the original wheel can be created.

https://files.pythonhosted.org/packages/7e/9d/517b12a42b0692c909ed348545114dae7d0b4014ef9075e18f6bf48834a1/PyQt6_Qt6-6.7.2-py3-none-macosx_11_0_arm64.whl

https://files.pythonhosted.org/packages/10/38/ba0313442c5e4327d52e6c48d2bb4b39099bf1d191bd872edfd8bb1392ef/PyQt6_Qt6-6.7.2-py3-none-macosx_10_14_x86_64.whl

- OS version: macOS 14.6.1
- Delocate version: 0.12.0

**Additional context**

Only files with the extensions
```
lib_exts=(".so", ".dylib", ".a")
```
are fused by `fuse_trees` but this Qt Framework binary has no extension at all.

Patching `fuse_trees` as follows works for those Qt packages, but is not be general enough to catch other cases.

```
def fuse_trees(to_tree, from_tree, lib_exts=(".so", ".dylib", ".a")):
...
        for fname in filenames:
            root, ext = splitext(fname)
            from_path = pjoin(from_dirpath, fname)
            to_path = pjoin(to_dirpath, fname)
            if not exists(to_path):
                _copyfile(from_path, to_path)
            elif cmp_contents(from_path, to_path):
                pass
            elif ext in lib_exts or fname.startswith('Qt'):
                # existing lib that needs fuse
                lipo_fuse(from_path, to_path, to_path)
...
```

HexDecimal:
Updates merging to automatically detect binary files and removes the older behavior. This handles the common case of binary files having no suffix. Fixes #228

Using `_is_macho_file` has false negatives. Instead, all files are passed to the `lipo` command and its return code is used to determine if the file was valid to be merged. This might have its own issues, but I'd need to completely rewrite `lipo_fuse` to make it graceful.

The `lib_exts` parameter of `fuse_trees` is now ignored and I intend for it to be removed completely later.

This reuses old tests instead of trying to keep the old behavior. Updating the tests now could conflict with PR #234.

### Pull Request Checklist

- [x] Read and follow the [CONTRIBUTING.md](https://github.com/matthew-brett/delocate/blob/master/CONTRIBUTING.md) guide
- [x] Mentioned relevant [issues](https://github.com/matthew-brett/delocate/issues)
- [x] Append public facing changes to [Changelog.md](https://github.com/matthew-brett/delocate/blob/master/Changelog.md)
- [x] Ensure new features are covered by tests
- [x] Ensure fixes are verified by tests

### DIFF
--- a/delocate/fuse.py
+++ b/delocate/fuse.py
@@ -15,7 +15,9 @@ libraries.
 from __future__ import annotations
 
 import os
+import re
 import shutil
+import subprocess
 import tempfile
 import warnings
 from collections.abc import Container
@@ -29,8 +31,8 @@ from .tools import (
     chmod_perms,
     cmp_contents,
     dir2zip,
-    lipo_fuse,
     open_rw,
+    replace_signature,
     zip2dir,
 )
 from .wheeltools import rewrite_record
@@ -80,6 +82,12 @@ def _retag_wheel(to_wheel: Path, from_wheel: Path, to_tree: Path) -> str:
     _update_wheelfile(to_tree, retag_name)
 
     return retag_name
+
+
+_RE_LIPO_UNKNOWN_FILE_STDERR = re.compile(
+    r"^fatal error: (?P<program>.+): "
+    r"can't figure out the architecture type of: (?P<file>.+)\n$"
+)
 
 
 def fuse_trees(
@@ -135,11 +143,28 @@ def fuse_trees(
             if cmp_contents(from_path, to_path):
                 continue
             try:
-                # Try to fuse this file with lipo
-                lipo_fuse(from_path, to_path, to_path)
-            except RuntimeError:
-                # existing not-lib file not identical to source
+                # Try to fuse this file using lipo
+                subprocess.run(
+                    [
+                        "lipo",
+                        "-create",
+                        from_path,
+                        to_path,
+                        "-output",
+                        to_path,
+                    ],
+                    check=True,
+                    text=True,
+                    capture_output=True,
+                )
+            except subprocess.CalledProcessError as exc:
+                if not _RE_LIPO_UNKNOWN_FILE_STDERR.match(exc.stderr):
+                    # Unexpected error on library file
+                    raise RuntimeError(exc.stderr) from None
+                # Existing non-library file not identical to source
                 _copyfile(from_path, to_path)
+            else:
+                replace_signature(to_path, "-")
 
 
 def fuse_wheels(
--- a/delocate/fuse.py
+++ b/delocate/fuse.py
@@ -17,9 +17,9 @@ from __future__ import annotations
 import os
 import shutil
 import tempfile
+import warnings
+from collections.abc import Container
 from os import PathLike
-from os.path import exists, relpath, splitext
-from os.path import join as pjoin
 from pathlib import Path
 
 from packaging.utils import parse_wheel_filename
@@ -83,17 +83,18 @@ def _retag_wheel(to_wheel: Path, from_wheel: Path, to_tree: Path) -> str:
 
 
 def fuse_trees(
-    to_tree: str | PathLike,
-    from_tree: str | PathLike,
-    lib_exts=(".so", ".dylib", ".a"),
-):
+    to_tree: str | PathLike[str],
+    from_tree: str | PathLike[str],
+    lib_exts: Container[str] | None = None,
+) -> None:
     """Fuse path `from_tree` into path `to_tree`.
 
-    For each file in `from_tree` - check for library file extension (in
-    `lib_exts` - if present, check if there is a file with matching relative
-    path in `to_tree`, if so, use :func:`delocate.tools.lipo_fuse` to fuse the
-    two libraries together and write into `to_tree`.  If any of these
-    conditions are not met, just copy the file from `from_tree` to `to_tree`.
+    Any files in `from_tree` which are not in `to_tree` will be copied over to
+    `to_tree`.
+
+    Files existing in both `from_tree` and `to_tree` will be parsed.
+    Binary files on the same path in both directories will be merged using
+    :func:`delocate.tools.lipo_fuse`.
 
     Parameters
     ----------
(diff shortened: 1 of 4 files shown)


delocate/fuse.py
```diff
@@ -102,30 +103,41 @@ def fuse_trees(
...
                 _copyfile(from_path, to_path)
-            elif cmp_contents(from_path, to_path):
-                pass
-            elif ext in lib_exts:
-                # existing lib that needs fuse
+                continue
+            if cmp_contents(from_path, to_path):
+                continue
+            try:
+                # Try to fuse this file with lipo
```
matthew-brett, line 138: OK - so you're leaving lipo-fuse to detect if this an executable binary on this platform?   Is this safe - I mean - does `lipo -create` guarantee to raise an error if one of the input files is not a mach-o binary?    I did a quick speedcheck - using `file` to check the filetype first for both files nearly doubles the time to do this check, compared to just calling `lipo -create`, but using the `python-magic` library, and parsing the output, adds virtually nothing to the check - and could also give a more informative message.  What do you think?

matthew-brett (commented):
First thoughts.

delocate/fuse.py
```diff
@@ -83,17 +83,18 @@ def _retag_wheel(to_wheel: Path, from_wheel: Path, to_tree: Path) -> str:
 
 
 def fuse_trees(
-    to_tree: str | PathLike,
-    from_tree: str | PathLike,
-    lib_exts=(".so", ".dylib", ".a"),
-):
+    to_tree: str | PathLike[str],
```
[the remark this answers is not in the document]
HexDecimal: Mypy's stricter settings don't like ambiguous generics.

`PathLike` is generic and `PathLike[str]` declares the parameter as expecting Unicode paths. It should generally be the same as string type it's unioned with. So `str  | PathLike[str]` compared to `str | bytes | PathLike[str] | PathLike[bytes]`. Even now the behavior of bytes types in paths be hard to follow.

delocate/fuse.py
HexDecimal: `lipo` returns error codes on non-library files. It will say the file isn't a library on stderr I think. I did some basic probing to make sure that it's at least more reliable than `otool` but I'm not a native MacOS dev. I'm unsure what to do other than to rely on the current tests.

`python-magic` sounds interesting but `fuse_trees` is not a function which does anything with errors other than ignoring them. A list if `lipo` error messages to ignore might be a better idea as I'm more worried about actual library files silently failing to fuse rather than files being falsely detected as libraries. `lipo` doesn't document its error codes very well so you might need to provide me with a list of which stderr strings should be suppressed.

HexDecimal: Also keep in mind the previous logic of all files with library-like prefixes being unconditionally fused using `lipo`.

matthew-brett: Happy to work on stderr - but I'm worried that processing stderr might be fragile across versions.   I wonder, should we raise an error when actual libraries fail to fuse, and propagate that error?  I guess we should decide if we want to propagate errors - for example - when the two libraries are the same architecture.

delocate/fuse.py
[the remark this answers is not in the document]
matthew-brett: Ah yes - sorry - but I was unclear how the type checker distinguishes byte-based Paths from string-based Paths.   When I'm running `mypy`, the type checker already detects type mismatch when making Paths from bytes.

delocate/fuse.py
HexDecimal: I'm unsure what that error would look like. Does `lipo` even fail in these cases?

I personally wouldn't fail here if the architectures are the same, especially if `lipo` allows that case.

matthew-brett: Lipo does not allow fusing two files with the same architecture, no:

```bash
$ cp ./delocate/tests/data/np-1.24.1_x86_64_random__sfc64.cpython-311-darwin.so foo.so
$ lipo -create delocate/tests/data/np-1.24.1_x86_64_random__sfc64.cpython-311-darwin.so foo.so -output bar.so
fatal error: /Library/Developer/CommandLineTools/usr/bin/lipo: delocate/tests/data/np-1.24.1_x86_64_random__sfc64.cpython-311-darwin.so and foo.so have the same architectures (x86_64) and can't be in the same fat output file
```

matthew-brett:
Relfecting more - the behavior change here is that any file that matches across the two archives, and that does not compare equal, will run through `lipo -create`.    Previously that would only have been true where files had suitable extensions. 

That's probably OK - but will now mean `lipo` can get a large range of file types, including large data files, and we're stress-testing `lipo`'s response to these files.   We can reasonably hope `lipo` will throw an error, but as we can now throw anything at it, it may segfault or similar.    That makes me think that it would be better to use `magic` to check the file type, given the cost is small - except for the extra dependency.

matthew-brett:
I guess checking with `magic` will also mean we can check whether we think `lipo -create` _should_ work, and then propagate the error if it does not.

HexDecimal:
Which MIME types should be expected from `python-magic`?

matthew-brett:
Here's some outputs:

```python
[ins] In [1]: import magic

[ins] In [2]: magic.from_file('README.rst')
Out[2]: 'Unicode text, UTF-8 text'

[ins] In [3]: magic.from_file('delocate/tests/data/np-1.24.1_x86_64_random__sfc64.cpython
         ...: -311-darwin.so')
Out[3]: 'Mach-O 64-bit x86_64 bundle, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL>'

[ins] In [4]: magic.from_file('delocate/tests/data/np-1.6.0_intel_lib__compiled_base.so')
         ...: 
Out[4]: 'Mach-O universal binary with 2 architectures: [x86_64:\\012- Mach-O 64-bit x86_64 bundle, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL>] [\\012- i386:\\012- Mach-O i386 bundle, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL>]'

[ins] In [5]: magic.from_file('delocate/tests/data/np-1.24.1_arm_random__sfc64.cpython-31
         ...: 1-darwin.so')
Out[5]: 'Mach-O 64-bit arm64 bundle, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL>'

[ins] In [6]: magic.from_file('wheel_makers/fakepkg_toplevel/module2.c')
Out[6]: 'C source, ASCII text'

[ins] In [7]: !dd if=/dev/random of=randbin bs=1 count=100
         ...: 
100+0 records in
100+0 records out
100 bytes transferred in 0.000585 secs (170940 bytes/sec)

[ins] In [8]: magic.from_file('randbin')
Out[8]: 'data'
```

HexDecimal:
I was asking for `mime=True` outputs, but I ran into issues trying to get them myself.

`python-magic` is not self contained and has dependencies on external binary files. This could make it hard to work with as an install dependency.

matthew-brett:
Yes, sorry, I sent the not-MIME outputs to show the level of detail they give.   I'll send the MIME outputs in a bit.

For the external dependency - given we're only running on Mac for the moment, and that binary packaging works currently for Mac - I think that's OK - I'm guessing they'll keep maintaining it.  But the CI will show us when they drop binary packaging.

HexDecimal:
The detailed output doesn't seem helpful in this situation. `application/x-mach-binary` is what I wanted. I don't know if `.a` files also fall other that MIME type.

> That's probably OK - but will now mean `lipo` can get a large range of file types, including large data files, and we're stress-testing `lipo`'s response to these files. We can reasonably hope `lipo` will throw an error, but as we can now throw anything at it, it may segfault or similar. That makes me think that it would be better to use `magic` to check the file type, given the cost is small - except for the extra dependency.

Could you demonstrate `lipo` behaving unreliably or slow? A `lipo` segfault seems like a big claim unless you're talking about deliberately malformed files which would pass a MIME type check anyways. Huge non-binary files should logically terminate early because they can't be parsed early.

matthew-brett:
Sure, segfaults and so on are unlikely - and probably it will be fine.   The main advantage of `magic` then will be to know when we're expecting `lipo` to work and raising the error when it doesn't.

HexDecimal:
There's only one expected error from `lipo` which is any "this is not a mach-o binary" error.

Example stderr:
```
fatal error: /Applications/Xcode_14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/lipo: can't figure out the architecture type of: /var/folders/9f/9p4dh6hs5yddrk7drxq8rc_80000gn/T/tmpv597xqb0/from_wheel/fakepkg2-1.0.dist-info/RECORD
```

I'll start by checking for this.

matthew-brett:
Looks reasonable to me ...

